PR #51, opened 2026-03-28
Improve test coverage for untested modules

DominicBurkart:
## Summary

Adds targeted unit tests for modules and code paths that had no test coverage, focusing on validation logic that can be exercised without live network access.

### New test files and what they cover

| File | Module under test | Key coverage added |
|---|---|---|
| `src/data/cache_tests.rs` | `InMemoryCache`, `CacheEntry` | Expiry (fresh/expired/negative TTL), insert/get/remove, custom TTL, cleanup of expired entries, clear, key overwrite |
| `tests/error_tests.rs` | `Error` enum | Exhaustive `mcp_error_code()` mapping (all variants negative), `error_type()` non-empty for all variants, `-32602` for all invalid-params variants, `Display` output for `RateLimited` with/without retry_after |
| `tests/mcp_types_tests.rs` | `mcp::types` | `GeographicBounds::contains` (inside/outside/edge), serde default deserialization for `FindNearbyStationsInput`, `SearchStationsByNameInput`, `GetStationByCodeInput`, `JsonRpcRequest` round-trip, `JsonRpcError::from` for param errors |
| `tests/handler_validation_tests.rs` | `McpToolHandler` | Input validation that fails **before** any API call: radius too large, limit exceeded, coordinates outside Paris, search query too short, journey origin/destination outside Paris |
| `tests/types_edge_cases_tests.rs` | `types.rs` | `BikeAvailability::saturating_add` overflow, default, single-type availability; `VelibStation::is_operational` without RT data, with Closed/Maintenance; `has_available_docks` with/without RT; `has_available_bikes` without RT and with filters; `StationReference::validate` for empty code/name/zero capacity/bad coords; serde round-trips for `StationStatus`, `DataFreshness`, `BikeTypeFilter`; `DataFreshness` boundary values (9.99/10.0/29.99/30.0/119.99/120.0 and negative); `distance_to` self=0 and symmetry; station validate capacity check |
| `tests/mcp_server_dispatch_tests.rs` | `McpServer` JSON-RPC | `tools/list` returns all 5 tools, `resources/list` returns 4 resources, unknown method returns error, unknown tool returns error, malformed JSON returns parse error (-32700), missing tool name returns error, response preserves request id, `/health` endpoint via `Server::router()` |

### Design decisions

- **No network required**: All new tests exercise pure logic, validation, caching, serde, and HTTP dispatch via axum `oneshot` -- none require the live Paris Open Data API.
- **Concise and expressive**: Each test validates one behavior with a descriptive name.
- **Boundary testing**: DataFreshness thresholds, geographic bounds edges, capacity overflow, and u16::MAX saturation.
- **Exhaustive variant coverage**: Error code/type mapping tests cover every `Error` variant to catch regressions when new variants are added.

### What is NOT changed

- Existing tests are untouched.
- No production code changes (only a `#[cfg(test)] mod cache_tests;` declaration added to `src/data/mod.rs`).

## Test plan

- [ ] `cargo test` passes (all new tests + existing tests)
- [ ] CI passes on the PR branch"


tests/handler_validation_tests.rs
DominicBurkart, line 1: **Merge conflict with PR #46.** Both this PR and #46 create `tests/handler_validation_tests.rs` with overlapping but different test coverage. Same applies to `tests/mcp_types_tests.rs`. Whichever merges second will conflict.

Recommendation: decide which PR's versions to keep, or consolidate into one PR. #46 additionally asserts on `error_type()` strings (stronger assertions), while this PR checks error message substrings (more brittle). Prefer #46's approach.

---

src/data/cache_tests.rs
DominicBurkart, line 1: **Overlap with PR #46.** PR #46 adds cache tests inline in `src/data/cache.rs` (using `mod tests`), while this PR creates a separate `src/data/cache_tests.rs`. The test cases are nearly identical (insert/get, expired entry, custom TTL, remove, cleanup, clear, overwrite). Only this PR's approach of a separate file differs structurally.

If #46 merges first, these become duplicate test coverage. Coordinate with #46 -- either drop this file or add only the tests that #46 does not cover (e.g., `tests/mcp_server_dispatch_tests.rs` and `tests/types_edge_cases_tests.rs` are unique to this PR and are valuable).

---

tests/handler_validation_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+        longitude: -74.006,
+        radius_meters: 500,
+        limit: 10,
+        availability_filter: None,
+    };
+    let err = handler().find_nearby_stations(input).await.unwrap_err();
+    assert!(
+        err.to_string().contains("coordinates")
+            || err.to_string().contains("Invalid")
+            || err.to_string().contains("service area"),
```
DominicBurkart: These assertions match on error message substrings (`contains("coordinates")`, `contains("Invalid")`, `contains("service area")`), which is fragile -- any rewording of the Display impl breaks the test. PR #46's version asserts on `err.error_type()` which is a stable programmatic identifier. Prefer that pattern.

---

tests/mcp_server_dispatch_tests.rs
```diff
@@ -0,0 +1,209 @@
...
+            "params": {}
+        }),
+    )
+    .await;
+
+    assert_eq!(status, StatusCode::OK);
+    let tools = body["result"]["tools"].as_array().unwrap();
+    assert_eq!(tools.len(), 5);
+
+    let names: Vec<&str> = tools.iter().map(|t| t["name"].as_str().unwrap()).collect();
```
DominicBurkart, line 51: Hardcoded count (`assert_eq!(tools.len(), 5)`) will break whenever a tool is added or removed. Consider asserting on a minimum count or checking that the expected names are present without asserting exact length. Same concern for `resources/list` (`assert_eq!(resources.len(), 4)`) on line 65.

---

DominicBurkart (commented):
**Major overlap with PR #46.** Both PRs create `tests/handler_validation_tests.rs` and `tests/mcp_types_tests.rs`, and both add cache tests (this PR in `src/data/cache_tests.rs`, #46 inline in `src/data/cache.rs`). Merging both will cause file-level conflicts or duplicate coverage.

Recommendation: Merge #46 first (it was opened earlier and has stronger assertions via `error_type()`), then rebase this PR and keep only the files unique to it:
- `tests/mcp_server_dispatch_tests.rs` (valuable -- exercises JSON-RPC dispatch via axum oneshot)
- `tests/types_edge_cases_tests.rs` (valuable -- boundary tests for DataFreshness, BikeAvailability overflow, distance symmetry)
- `tests/error_tests.rs` (partially overlaps with #46's inline error tests, but adds the exhaustive `mcp_error_code` negativity check)

Also note: hardcoded tool/resource counts in `mcp_server_dispatch_tests.rs` will break when tools are added. Consider asserting on minimum counts or just checking expected names are present.

---

DominicBurkart (commented):
[Automated review by Claude]

## Summary

Targeted unit tests across cache, error, MCP types/dispatch, handler validation, and `types.rs` edge cases. All claimed modules are exercised. Tests are pure-logic and network-free, deterministic in the common case, and the assertions are meaningful (not coverage padding).

## Strengths

- Good boundary coverage: `DataFreshness::from_age` at 9.99/10/29.99/30/119.99/120, `BikeAvailability::saturating_add` at `u16::MAX`, `GeographicBounds` edge inclusivity.
- Exhaustive enumeration of `Error` variants for `mcp_error_code`/`error_type`/`-32602` mapping is a robust regression guard against new variants.
- `mcp_server_dispatch_tests.rs` exercises real routing via `axum::oneshot` rather than mocking the dispatcher — high-value, low-bloat.
- No excessive mocking; tests construct domain types directly.

## Issues

1. **Brittle hardcoded counts** in `mcp_server_dispatch_tests.rs`: `tools.len() == 5` and `resources.len() == 4` will break the moment a tool or resource is added. Prefer `assert!(tools.len() >= 5)` plus the existing per-name asserts (this was also flagged by the author's self-review).

2. **`tools_call_missing_name_returns_error` codifies sub-optimal behavior**: it asserts HTTP 500 + plain `{"error": "..."}` body. The MCP-correct behavior is a JSON-RPC error object. The test comment acknowledges this. Consider filing a follow-up issue and marking the test `#[ignore]` or rewriting once the server is fixed, so the test doesn't lock in the bug.

3. **Minor flakiness risk**: `cache_entry_expired_with_negative_ttl` and `expired_entry_returns_none` rely on `Utc::now() > expires_at` where `expires_at = Utc::now() + Duration::seconds(-1)`. The 1-second margin is fine in practice but `Duration::milliseconds(-1)` would be tight. Current values are safe.

4. **PR overlap and merge conflict**: `mergeable_state` is `dirty` and there is significant overlap with PR #46 (already noted in the author's own review). Needs rebase after #46 lands; keep only `mcp_server_dispatch_tests.rs`, `types_edge_cases_tests.rs`, and the exhaustive variant tests in `error_tests.rs`.

5. **`search_by_name_rejects_short_query`** asserts `"too short"` substring — but the handler returns `Error::Internal(anyhow!("Search query too short"))`, which is a generic `Internal` error rather than `Validation`. Test passes, but the production code should arguably return `Error::Validation` so the JSON-RPC code becomes `-32602` not `-32603`. Worth a follow-up.

## Verdict

Net positive. Ship after rebasing on #46 and trimming hardcoded counts. The two behavior-locking tests (#2, #5) should be tracked as follow-ups so they don't entrench current bugs.

---

tests/handler_validation_tests.rs
DominicBurkart: Addressed. `tests/handler_validation_tests.rs` and `tests/mcp_types_tests.rs` have been replaced with the versions from PR #46 (already on `main`), which use `error_type()` string assertions. The `src/data/cache_tests.rs` file and its `mod cache_tests` declaration in `src/data/mod.rs` have been removed since the inline tests from PR #46 already live in `src/data/cache.rs`. The PR now retains only the files unique to it: `tests/mcp_server_dispatch_tests.rs`, `tests/types_edge_cases_tests.rs`, and `tests/error_tests.rs`.

---

src/data/cache_tests.rs
DominicBurkart: Addressed. `src/data/cache_tests.rs` has been deleted and the `#[cfg(test)] mod cache_tests;` declaration removed from `src/data/mod.rs`. PR #46's inline tests in `src/data/cache.rs` are already on `main` and cover the same cases.

---

tests/handler_validation_tests.rs
DominicBurkart: Addressed. `tests/handler_validation_tests.rs` has been replaced with the version from PR #46, which asserts on `err.error_type()` (e.g. `"search_radius_too_large"`, `"result_limit_exceeded"`, `"invalid_coordinates"`) rather than matching against error message substrings.

---

tests/mcp_server_dispatch_tests.rs
DominicBurkart: Addressed. The hardcoded counts have been replaced with minimum-count assertions: `assert!(tools.len() >= 5)` and `assert!(resources.len() >= 4)`. The per-name presence checks for all five known tools are retained. The test function names have also been updated (`tools_list_returns_expected_tools`, `resources_list_returns_resources`) to reflect the looser assertion.

---

DominicBurkart (commented):
[Automated review by Claude — HEAD 703655f4, 2026-04-17]

The previous automated review (2026-04-16) was against commit `78f4643`. Two new commits have landed since then (`ac88e73`, `703655f4`) that address the prior review's major concerns. This review covers the current state of the diff.

## What changed since the last review

- `src/data/cache_tests.rs` dropped (duplicate of inline tests already in `cache.rs` from PR #46 — correct call).
- The `#[cfg(test)] mod cache_tests;` declaration in `src/data/mod.rs` is consequently removed from the diff.
- Hardcoded `tools.len() == 5` / `resources.len() == 4` replaced with `>= 5` / `>= 4` plus named-tool assertions.

The four remaining test files are: `tests/error_tests.rs`, `tests/handler_validation_tests.rs`, `tests/mcp_server_dispatch_tests.rs`, `tests/mcp_types_tests.rs`, and `tests/types_edge_cases_tests.rs`.

## Analysis against the review checklist

### Cache tests / TTL math
No cache tests remain in this PR (dropped in favour of PR #46's inline tests). Nothing to flag here.

### Error variant exhaustiveness
`all_error_variants_have_valid_mcp_codes` and `all_error_variants_have_nonempty_error_type` enumerate every `Error` variant by hand. This works now, but it won't self-enforce when new variants are added — the test will still compile and pass while silently under-covering the new variant. A macro-based approach (or a compile-time exhaustive match guard) would be stronger, but the current approach is an acceptable trade-off for a test suite. **Minor concern**: `StationNotFound` is present in `all_error_variants_have_valid_mcp_codes` but absent from `invalid_params_errors_use_correct_code` — if `StationNotFound` maps to `-32602`, that exclusion is a gap; if it maps to something else (e.g. `-32001`), the omission is intentional and fine. Worth a comment in the test explaining why it is excluded.

### Handler validation — early rejection without API calls
The tests construct `McpToolHandler::new()` and call the async handler methods. For variants that fail validation *before* any network call (radius, limit, coordinates, short query), this is correct and sufficient. One edge case: `find_nearby_rejects_coordinates_outside_service_area` explicitly allows the test to succeed even if the network call fires and returns data:

```rust
if let Err(err) = &result {
    let err_type = err.error_type();
    assert!(
        err_type == "outside_service_area" || err_type == "http_error",
        ...
    );
}
// If it succeeded, the coordinates were within 50km -- also acceptable.
```

This test gives zero guarantees — it passes unconditionally. Either pick coordinates that are provably > 50 km from Paris city hall and assert `outside_service_area`, or delete the test. As written it is noise.

### MCP server dispatch — real axum routing vs stub
`mcp_server_dispatch_tests.rs` uses `axum::oneshot` against `McpServer::new().router()`. This is real routing — not mocked. Good.

### All 5 tools covered in dispatch test
`tools_list_returns_expected_tools` checks for all five by name:
`find_nearby_stations`, `get_station_by_code`, `search_stations_by_name`, `get_area_statistics`, `plan_bike_journey`. That is complete for the currently declared tools.

### Weak assertions
- `search_stations_rejects_short_query` asserts only `result.is_err()` — it does not check `error_type()`. Given that the underlying implementation returns `Error::Internal` (not `Error::Validation`) for this case, the missing `error_type` check is arguably hiding a semantic bug in production code. The test is still useful as a regression guard but should assert the exact error type so a future fix to return `Error::Validation` is forced to update the test.
- `tools_call_missing_name_returns_error` asserts HTTP 500 + plain `{"error": "..."}`. This locks in non-conformant MCP behavior. The comment in the test acknowledges this. A `#[ignore]` tag with a `// TODO: fix server to return JSON-RPC error` comment would prevent it from hiding a known bug while keeping it as documentation.

### Test pollution / shared state
All tests construct a fresh `McpToolHandler::new()` or `McpServer::new()`. No shared mutable state is visible. No `lazy_static` or `once_cell` globals. No pollution risk.

### `mcp_server_dispatch_tests.rs` — hardcoded counts
Addressed: `>= 5` and `>= 4` with explicit name checks. Good.

## Issues (priority order)

**Medium**

1. **`find_nearby_rejects_coordinates_outside_service_area` is a no-op test.** It passes whether validation fires or not, and allows `"http_error"` as an acceptable outcome (meaning it would pass even if the test makes a real live API call that succeeds). Either pin coordinates that are provably outside 50 km from the city centre and assert `outside_service_area` unconditionally, or remove the test.

2. **`tools_call_missing_name_returns_error` encodes a known protocol bug.** Asserting HTTP 500 with a plain string body locks in non-conformant behaviour. Add `#[ignore]` and a comment pointing to a follow-up issue so it doesn't silently prevent the bug from being fixed.

**Low**

3. **`search_stations_rejects_short_query` has a weak assertion.** It only checks `is_err()`. Add `assert_eq!(result.unwrap_err().error_type(), "internal")` (to match current behaviour) or fix the production code to return `Error::Validation` and then assert `"validation"`. Either way, make the assertion explicit.

4. **`StationNotFound` exclusion from `invalid_params_errors_use_correct_code` is undocumented.** Add a one-line comment explaining why it is intentionally absent (or include it if `-32602` is its code).

5. **`rate_limited_display_with_and_without_retry`** asserts `!msg.contains("retry after")` for the `None` case. This is a fragile negative assertion — if the message is ever rephrased to say "retry_after" or "retry-after", the test will silently pass while the user-visible message has regressed. Prefer asserting on the exact expected string.

## Verdict

Net positive change. The prior review's two main concerns (duplicate cache tests, hardcoded counts) are resolved. The remaining issues are minor and pre-existing. The axum `oneshot` dispatch tests and the exhaustive `DataFreshness`/`BikeAvailability` boundary tests are the highest-value additions. Address issue #1 (no-op test) before merge; issues #2–5 can be follow-ups.

---
Session: https://claude.ai/code/session_01UMqGrE2sV8UmCCz3NLEs9i

---

DominicBurkart:
Automated triage pass complete: a fresh code review was posted finding one issue requiring a fix before merge — `find_nearby_rejects_coordinates_outside_service_area` passes unconditionally since it allows both success and any error response (rewrite with coordinates provably outside the 50km service area and an unconditional `outside_service_area` assertion); all DominicBurkart review comments were confirmed already addressed in prior commits; and all 7 CI checks are passing on the current head. Fix the no-op test, then this PR is ready for human review.

Session: https://claude.ai/code/session_01UMqGrE2sV8UmCCz3NLEs9i

---

DominicBurkart:
Addressed the no-op `find_nearby_rejects_coordinates_outside_service_area` in 8be094e.

Investigation: the `OutsideServiceArea` branch is not reachable via `McpToolHandler::find_nearby_stations` because `is_valid_paris_metro` (bounding box lat 48.7-49.0, lon 2.0-2.6) is checked first, and that box is fully contained within the 50km service radius from Paris City Hall -- any point inside the box is at most ~31km away. So there are no coordinates that fail service-area while passing metro-box.

Fix: renamed the test to `is_within_paris_service_area_rejects_distant_coordinates` and rewrote it as a direct, unconditional assertion against `Coordinates::is_within_paris_service_area()` using Reims (~130km) and Lyon (~390km) as provably-outside points, plus Paris City Hall as an inside sentinel. Ran `cargo check --tests` and `cargo test --test handler_validation_tests` locally -- all 8 tests pass.

If the intent is to exercise the handler's service-area branch end-to-end, the production-code ordering would need to change (e.g. swap the checks, or widen the metro box). Happy to do that in a follow-up if desired.

---

DominicBurkart (commented):
[Automated review by Claude — HEAD `8be094e`, 2026-04-20]

This review covers only the delta since the last automated review (`703655f4`, 2026-04-19). One commit landed since then.

## What changed

The sole blocker from the previous review has been addressed:

- **`find_nearby_rejects_coordinates_outside_service_area` (no-op) removed.** Replaced by `is_within_paris_service_area_rejects_distant_coordinates`, which directly calls `Coordinates::is_within_paris_service_area()` with three unconditional assertions:
  - Reims (~130 km ENE) → must return `false` ✓
  - Lyon (~390 km SSE) → must return `false` ✓
  - Paris City Hall (48.8565, 2.3514) → must return `true` ✓

The fix is correct. The test no longer passes unconditionally; it will catch any regression in the service-area guard. The test comment accurately explains why the `OutsideServiceArea` handler branch is not directly reachable via `McpToolHandler::find_nearby_stations` (the `is_valid_paris_metro` bounding-box check fires first and the box is fully inside the 50 km radius).

## Remaining open items (carry-overs, all low-priority follow-ups)

These were flagged in the 2026-04-17 review and explicitly deferred — nothing new here:

1. **`tools_call_missing_name_returns_error` asserts HTTP 500 + plain string** — locks in non-conformant MCP behaviour. Consider `#[ignore]` + follow-up issue.
2. **`search_stations_rejects_short_query` asserts only `is_err()`** — weak; doesn't pin the error type. Acceptable as a regression guard but won't catch a semantics change from `Internal` to `Validation`.
3. **`StationNotFound` absent from `invalid_params_errors_use_correct_code`** — add a comment explaining why (intentional omission or oversight).
4. **`rate_limited_display_with_and_without_retry` uses a negative substring assertion** (`!msg.contains("retry after")`). Prefer asserting the exact expected string.

None of these block merge.

## Verdict

The blocking issue from the previous review is resolved. All four remaining items are pre-existing low-priority follow-ups. This PR is ready for merge once the draft status is cleared.

---
Session: https://claude.ai/code/session_01FPKocTFe3534k7H3bKxG4Z

---

DominicBurkart:
All DominicBurkart review comments are confirmed addressed in the current branch:

1. **handler_validation_tests.rs fragile assertions** (threads 1 & 3): Already fixed — the branch uses `error_type()` assertions throughout, and the no-op `find_nearby_rejects_coordinates_outside_service_area` was replaced with `is_within_paris_service_area_rejects_distant_coordinates`, which directly tests `Coordinates::is_within_paris_service_area()` with unconditional assertions (Reims, Lyon as outside; Paris City Hall as inside).

2. **src/data/cache_tests.rs overlap** (thread 2, outdated): Already resolved — the file was removed from the branch; `src/data/mod.rs` has no `mod cache_tests` declaration.

3. **mcp_server_dispatch_tests.rs hardcoded counts** (thread 4): Already fixed — `assert!(tools.len() >= 5)` and `assert!(resources.len() >= 4)` replace the exact-count assertions; named tool presence checks are retained.

**Remaining blocker — merge conflict**: The PR cannot be auto-merged because `tests/handler_validation_tests.rs` and `tests/mcp_types_tests.rs` were added independently on both this branch and `main` (via PR #46), creating an add/add conflict at the git history level. This cannot be resolved through the GitHub API alone.

To resolve locally:
```sh
git checkout improve-test-coverage-2
git fetch origin
git rebase origin/main
# For tests/handler_validation_tests.rs: keep the branch version
# (is_within_paris_service_area_rejects_distant_coordinates is better)
# For tests/mcp_types_tests.rs: keep either — they are identical
git push --force-with-lease origin improve-test-coverage-2
```

After rebasing, this PR should be mergeable with only the 3 truly unique files as additions: `tests/error_tests.rs`, `tests/mcp_server_dispatch_tests.rs`, and `tests/types_edge_cases_tests.rs`.

---

DominicBurkart (commented):
[Automated review by Claude — HEAD `b98ad44`, 2026-04-21]

This review covers the delta since the last automated review (`8be094e`, 2026-04-20). Two commits have landed since (`3b11a06`, `b98ad44`), both resolving merge conflicts with `main` after PR #46 landed. No new test logic was introduced; the existing improved test bodies were preserved through the merge.

## What changed since the last review

- Merge conflicts with `main` resolved in `handler_validation_tests.rs` and `mcp_types_tests.rs`.
- The improved `is_within_paris_service_area_rejects_distant_coordinates` test (Reims / Lyon / City Hall) is retained as the stronger of the two conflicting versions. Verified in the current diff (lines 51–82 of `handler_validation_tests.rs`).
- `mcp_types_tests.rs` matches main post-merge (no semantic change in this PR beyond the initial addition).
- `mergeable_state` is no longer `dirty`.

## Checklist re-evaluation against HEAD

- **Tests exercise behavior, not just coverage?** Yes. `DataFreshness` boundary tests (9.99/10.0/29.99/30.0/119.99/120.0), `BikeAvailability::saturating_add` at `u16::MAX`, `GeographicBounds` inclusive-edge assertions, and exhaustive `Error` variant enumeration all pin real semantics.
- **Over-mocking?** No. Dispatch tests use `axum::oneshot` against the real `McpServer::new().router()` — no stubbed dispatcher. Handler validation tests construct `McpToolHandler::new()` directly.
- **Flaky patterns?** No `sleep`, no wall-clock dependencies, no network. `Utc::now()` is used only to populate `last_update` fields that are never asserted against.
- **Property tests / fuzzing?** Absent, but the enumerated-variant and boundary-value approach is sufficient for the surface under test. Not a blocker.
- **Production code changed?** No. Diff is tests-only (`src/data/mod.rs` `#[cfg(test)] mod cache_tests;` declaration was removed when `cache_tests.rs` was dropped — confirmed net-zero production changes in the current diff).
- **Bloat?** Acceptable. 952 additions across 5 files, each test is single-behavior and short.
- **Secrets/PII?** None. Fixtures use Paris landmark coordinates, New York, Reims, Lyon — all public.

## Carry-over low-priority follow-ups (from 2026-04-17 / 2026-04-20 reviews)

None block merge. All were acknowledged and explicitly deferred:

1. `tools_call_missing_name_returns_error` codifies the HTTP 500 + plain-string-body bug. Consider `#[ignore]` + follow-up issue so a fix isn't blocked by a passing test.
2. `search_stations_rejects_short_query` asserts only `is_err()` — pin the `error_type` to force an explicit update when the production `Error::Internal` → `Error::Validation` fix lands.
3. `StationNotFound` is present in `all_error_variants_have_valid_mcp_codes` but absent from `invalid_params_errors_use_correct_code`. Add a one-line comment clarifying whether this is intentional (non-`-32602` code) or an oversight.
4. `rate_limited_display_with_and_without_retry` uses a negative substring assertion (`!msg.contains("retry after")`). Prefer asserting the exact expected string so rephrasing doesn't silently regress the user-visible message.

## Verdict

Ready to merge once the draft flag is cleared. The merge-conflict resolution preserved the strongest version of the one previously-blocking test. All remaining items are pre-existing low-priority follow-ups, not new regressions.

---
Session: https://claude.ai/code/session_01BjjYHHSeWpWs4eut1FeegC

---

DominicBurkart:
**Automated CRON status (review + address-comments + CI fanout):** Code-review subagent deduped (existing owner review on head from 2026-04-21 concluded "Ready to merge once the draft flag is cleared"). CI-fix subagent cleared the `add/add` merge conflict in `tests/handler_validation_tests.rs` by aligning it with main's version, ran `update_pull_request_branch` (base fast-forwarded to main `f411180`), and committed `ac6777d` reintroducing the HEAD-only `is_within_paris_service_area_rejects_distant_coordinates` test so no coverage is lost; locally verified `cargo test --lib` 67/67 pass, fmt + clippy clean. Address-comments subagent confirmed all 4 threads are auto-resolved (3 outdated by the merge + 1 replied "addressed"); **all 9 CI checks now green** (Build, Code Coverage, Test Suite, Cargo Sort, Security Audit, Clippy, Rustfmt, Feature Testing, All Checks Passed). **Ready for human review — flipping draft flag off.** — https://claude.ai/code/session_012Am9fyTTsd6FPjs374UBW5

---

PR #52, opened 2026-03-31
Extract station filtering helper

DominicBurkart:
## Summary
- Extract `find_stations_within_radius` helper from repeated filtering logic in `find_nearby_stations()` and `plan_bike_journey()`
- Consolidates distance calculation, filtering, sorting, and truncation into a single reusable function
- Net: -25 lines, eliminates three instances of duplicate filtering logic

## Test plan
- [ ] `cargo test --all-features` passes (all 34+ tests)

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -25,6 +25,13 @@ const PARIS_CITY_HALL: Coordinates = Coordinates {
 
 /// Find stations near a point, filtering by distance and a custom predicate,
 /// sorted by distance and truncated to `limit` results.
+///
+/// Each matched station is cloned into the returned `StationWithDistance`. This
+/// is intentional: callers that hold the original slice (e.g. `plan_bike_journey`,
+/// which needs `all_stations` for both the pickup and the dropoff pass) must not
+/// have their data consumed. For `find_nearby_stations`, which previously used
+/// `into_iter()`, this introduces one extra clone per matched station; given the
+/// Paris Velib network of ~1,400 stations this overhead is negligible.
 fn find_stations_within_radius(
     stations: &[VelibStation],
     origin: &Coordinates,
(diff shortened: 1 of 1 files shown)


src/mcp/handlers.rs
```diff
@@ -23,6 +23,35 @@ const PARIS_CITY_HALL: Coordinates = Coordinates {
     longitude: 2.3514,
 };
 
+/// Find stations near a point, filtering by distance and a custom predicate,
+/// sorted by distance and truncated to `limit` results.
+fn find_stations_within_radius(
```
DominicBurkart, line 35: **Subtle semantic change: `into_iter()` to `.iter()`.** The original `find_nearby_stations` used `all_stations.into_iter()` which consumed the vec, avoiding clones. The new `find_stations_within_radius` takes `&[VelibStation]` and calls `station.clone()` inside the helper. This is correct and actually aligns with how `plan_bike_journey` already worked (it called `.iter()` since it needed `all_stations` for both pickup and dropoff). Net effect: `find_nearby_stations` now does one extra clone per matched station. This is negligible for the dataset size (Paris has ~1,400 Velib stations) but worth noting for documentation.

---

src/mcp/handlers.rs
```diff
@@ -23,6 +23,35 @@ const PARIS_CITY_HALL: Coordinates = Coordinates {
...
+                })
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    results.sort_by_key(|s| s.straight_line_distance_meters);
+    results.truncate(limit);
+    results
```
DominicBurkart, line 59: **Consider adding a unit test for the extracted helper.** `find_stations_within_radius` is now a standalone function with a clear contract (filter by radius, apply predicate, sort by distance, truncate to limit). This is an ideal candidate for a focused unit test with mock station data -- verifying filtering, sort order, truncation, and predicate behavior. Since this is a refactor PR, you could add this in a follow-up, but it would strengthen the extraction.

---

DominicBurkart (commented):
## Review: Extract station filtering helper

**Verdict: Clean, correct refactoring. LGTM -- ready for human review.**

### What works well
- `find_stations_within_radius` cleanly consolidates 3 instances of duplicated filter-sort-truncate logic into one reusable function.
- The `impl Fn(&VelibStation) -> bool` predicate parameter is idiomatic and flexible.
- Net -25 lines with no behavioral change.
- No conflicts with other open PRs.

### Minor notes
1. **Subtle clone addition**: `find_nearby_stations` previously consumed `all_stations` via `into_iter()`, avoiding clones. The new helper takes `&[VelibStation]` and clones matches. This is negligible for ~1,400 stations but worth documenting.
2. **Unit test opportunity**: The extracted helper has a clean contract ideal for direct unit testing (filter, sort, truncate, predicate). Consider a follow-up test PR.

### Bugs found
None. The refactoring is behavior-preserving.

### Secrets/PII
None.

### CI
All 6 checks pass (Build, Test Suite, Feature Testing, Clippy, Rustfmt, Security Audit).

---
_Generated by [Claude Code](https://claude.ai/code) review bot_

---

src/mcp/handlers.rs
DominicBurkart: Good catch — addressed in the latest commit. The doc comment on `find_stations_within_radius` now explicitly documents the clone-per-matched-station behaviour and explains why the helper takes `&[VelibStation]` instead of consuming the vec (so callers like `plan_bike_journey` can reuse `all_stations` for both pickup and dropoff passes). The performance note about ~1,400 Velib stations making this negligible is included too.

---

src/mcp/handlers.rs
DominicBurkart: Done in the same commit — added a `#[cfg(test)] mod tests` block directly in `src/mcp/handlers.rs` with 9 unit tests covering all the contract points you listed:\n\n- **Filtering by radius**: stations outside the radius are excluded; an all-distant set returns empty.\n- **Sort order**: results are returned in ascending distance order regardless of input order.\n- **Truncation**: `limit` is applied after sorting, so only the closest N stations are kept; a limit larger than the match count returns all matches.\n- **Predicate behaviour**: only stations satisfying the predicate are included; a predicate that rejects everything returns empty.\n- **Combined**: a single test exercises radius + predicate + truncation together to guard against regressions in the interaction between all three.

---

PR #58, opened 2026-04-03
arch: extract fetch helper, consolidate coordinate validation, centralize constants

DominicBurkart:
## Summary

Three small architecture improvements targeting duplication and single-source-of-truth violations identified in the handler and types layers.

---

### Change A — `McpToolHandler::fetch_all_stations` helper (`src/mcp/handlers.rs`)

**Problem:** Four handler methods (`find_nearby_stations`, `search_stations_by_name`, `get_area_statistics`, `plan_bike_journey`) each repeated the same two-line pattern to acquire a write lock and fetch live station data:

```rust
let mut data_client = self.data_client.write().await;
let all_stations = data_client.get_all_stations(true).await?;
```

**Fix:** Extracted a private async helper:

```rust
async fn fetch_all_stations(&self) -> Result<Vec<VelibStation>> {
    let mut data_client = self.data_client.write().await;
    data_client.get_all_stations(true).await
}
```

All four call sites now read `self.fetch_all_stations().await?`. Any future change to how live data is fetched (e.g. adding tracing, switching lock strategy) is a single-line edit instead of a four-site hunt.

---

### Change B — `Coordinates::validate_for_velib_service` (`src/types.rs`)

**Problem:** `find_nearby_stations` and `plan_bike_journey` both duplicated a two-step coordinate check: first `is_valid_paris_metro()` → `InvalidCoordinates`, then `is_within_paris_service_area()` → `OutsideServiceArea`. `plan_bike_journey` ran this block twice (once for origin, once for destination), meaning the pattern appeared three times across the two handlers.

**Fix:** Added a method on `Coordinates` that encapsulates the ordered pair of checks and returns the appropriate `crate::Error` variant:

```rust
pub fn validate_for_velib_service(&self) -> Result<(), crate::Error> {
    if !self.is_valid_paris_metro() {
        return Err(crate::Error::InvalidCoordinates { .. });
    }
    if !self.is_within_paris_service_area() {
        return Err(crate::Error::OutsideServiceArea { .. });
    }
    Ok(())
}
```

Call sites collapse to a single `coord.validate_for_velib_service()?`. A test covering the new method is added to `types.rs`.

---

### Change C — Constants audit (`src/mcp/handlers.rs`, `src/data/client.rs`)

**Finding:** No deduplication was necessary.

- `MAX_SEARCH_RADIUS` and `MAX_RESULT_LIMIT` are already single module-level constants in `handlers.rs` with no copies elsewhere.
- `REFERENCE_CACHE_TTL_MINUTES` and `REALTIME_CACHE_TTL_MINUTES` are already single module-level constants in `client.rs` with no copies elsewhere.

The `PARIS_CITY_HALL` constant that was previously defined at module level in `handlers.rs` (and also inline inside `is_within_paris_service_area` in `types.rs`) is now only used inside `validate_for_velib_service`, so the module-level copy in `handlers.rs` has been removed — one fewer definition.

---

## Test plan

- [ ] `cargo build` passes with no new warnings
- [ ] `cargo test` passes (existing tests unchanged; new `test_validate_for_velib_service` test added)
- [ ] Manual smoke-test: call `find_nearby_stations` with out-of-bounds coordinates and confirm `InvalidCoordinates` / `OutsideServiceArea` errors still surface correctly
- [ ] Manual smoke-test: call `plan_bike_journey` with invalid origin or destination and confirm correct error is returned for each

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -17,6 +17,114 @@ use tokio::sync::RwLock;
 const MAX_SEARCH_RADIUS: u32 = 5000; // 5km
 const MAX_RESULT_LIMIT: u16 = 100;
 
+/// Validate that a coordinate is within the Velib service area, returning the
+/// appropriate `Error` if not. Centralizes the two checks previously duplicated
+/// across each handler (valid Paris metro bounds + 50km-of-City-Hall).
+fn ensure_in_service_area(coords: &Coordinates) -> Result<()> {
+    if !coords.is_valid_paris_metro() {
+        return Err(Error::InvalidCoordinates {
+            latitude: coords.latitude,
+            longitude: coords.longitude,
+        });
+    }
+    if !coords.is_within_paris_service_area() {
+        return Err(Error::OutsideServiceArea {
+            distance_km: coords.distance_to_paris_city_hall_km(),
+        });
+    }
+    Ok(())
+}
+
+/// Aggregate a set of stations into area statistics.
+///
+/// Pure function over an iterator so it can be unit-tested without any
+/// data-client wiring. Stations without real-time data contribute to
+/// `total_stations`, `operational_stations` (per `is_operational`), and
+/// `total_capacity`, but their bike/dock counts are treated as 0 -- the data is
+/// genuinely absent, and invented zeros for `has_bikes`/`has_docks` would be
+/// misleading. The `occupancy_rate` is bikes-over-capacity and returns 0.0
+/// when no capacity is present.
+fn aggregate_area_statistics<'a, I>(stations: I) -> AreaStatistics
+where
+    I: IntoIterator<Item = &'a VelibStation>,
+{
+    let mut total_stations = 0u32;
+    let mut operational_stations = 0u32;
+    let mut total_capacity = 0u32;
+    let mut total_mechanical = 0u32;
+    let mut total_electric = 0u32;
+    let mut total_available_docks = 0u32;
+
+    for station in stations {
+        total_stations += 1;
+        if station.is_operational() {
+            operational_stations += 1;
+        }
+        total_capacity += u32::from(station.reference.capacity);
+        if let Some(rt) = &station.real_time {
+            total_mechanical += u32::from(rt.bikes.mechanical);
+            total_electric += u32::from(rt.bikes.electric);
+            total_available_docks += u32::from(rt.available_docks);
+        }
+    }
+
+    let total_bikes = total_mechanical + total_electric;
+    let occupancy_rate = if total_capacity > 0 {
+        f64::from(total_bikes) / f64::from(total_capacity)
+    } else {
+        0.0
+    };
+
+    AreaStatistics {
+        total_stations,
+        operational_stations,
+        total_capacity,
+        available_bikes: AvailableBikesStats {
+            mechanical: total_mechanical,
+            electric: total_electric,
+            total: total_bikes,
+        },
+        available_docks: total_available_docks,
+        occupancy_rate,
+    }
+}
+
+/// Find stations near a point, filtering by distance and a custom predicate,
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
Refactor looks correct overall — no behavioral changes, no deadlock risk, constant removal is safe. One substantive finding: the third test case in `test_validate_for_velib_service` does not exercise the code path it claims to cover.

## `OutsideServiceArea` arm is never tested

`src/types.rs`, `test_validate_for_velib_service`, third case:

```rust
// Within bounding box but outside 50km service radius
// (no real coordinate satisfies this given the tight bbox, but the
//  method ordering is tested: metro check fires first)
let very_far = Coordinates::new(50.0, 2.3514);
assert!(very_far.validate_for_velib_service().is_err());
```

`lat=50.0` is outside the Paris metro bounding box, so `is_valid_paris_metro()` fires and the function returns `InvalidCoordinates` — the `is_within_paris_service_area()` branch in `validate_for_velib_service` is never reached. The comment acknowledges this ("metro check fires first") but draws the wrong conclusion: testing that the ordering is correct does not substitute for testing the second branch.

The `OutsideServiceArea` error variant from `validate_for_velib_service` has zero test coverage. If someone later changes the second `if` block (wrong error variant, wrong distance calculation, wrong guard condition) no test will catch it.

**Fix:** call `is_within_paris_service_area` directly to find the threshold, pick a coordinate that passes `is_valid_paris_metro` but fails `is_within_paris_service_area`, and assert the exact variant:

```rust
// A point inside the metro bbox but > 50 km from City Hall.
// The metro bbox extends to ~49.1 N; pick something near the northern edge
// that is still within bbox lat/lon bounds but far enough from City Hall.
// Adjust the specific values to whatever is geometrically valid for your bbox.
let edge_of_bbox = Coordinates::new(49.05, 2.3514);
assert!(matches!(
    edge_of_bbox.validate_for_velib_service(),
    Err(crate::Error::OutsideServiceArea { .. })
));
```

If no coordinate satisfies "inside metro bbox AND outside 50 km radius" (which the comment suggests may be true), then `validate_for_velib_service` contains a dead branch. That would mean `is_within_paris_service_area` is always redundant after `is_valid_paris_metro`, and both the branch and the `OutsideServiceArea` error path should be removed or the bounding box bounds should be re-examined.

Either way, the test should be corrected so it verifies what it states.

---

Everything else checks out:

- `fetch_all_stations` — write-lock acquisition is fine; `&self` holds no prior lock, Tokio's `RwLock` is not reentrant, and the guard is released before the returned `Vec` is used by callers. No deadlock risk.
- `validate_for_velib_service` check ordering — `is_valid_paris_metro` before `is_within_paris_service_area`, matching the original inline ordering in both handlers.
- `PARIS_CITY_HALL` removal from `handlers.rs` — safe; the only remaining use is the `const` local inside `validate_for_velib_service` in `types.rs`, value is identical (`48.8565, 2.3514`).
- Pure refactor — no observable behavioral changes.

https://claude.ai/code/session_01G7ndAwj9iNcWH8D6PyBBHB

---

DominicBurkart:
**Cron status — all 3 subagents complete; closing as redundant.** Code review posted one finding (OutsideServiceArea branch has zero test coverage); no review threads to address; CI investigation found that `main` has already incorporated all three of this PR's changes in evolved/improved form: (A) `fetch_all_stations` helper was tried and reverted in favor of pure `find_stations_within_radius`/`aggregate_area_statistics` functions, (B) `validate_for_velib_service` is superseded by `ensure_in_service_area`, (C) `PARIS_CITY_HALL` removal already done and further improved with public constants in `types.rs`. This PR is redundant — merging it would be a no-op against current `main`. Closing.

https://claude.ai/code/session_01G7ndAwj9iNcWH8D6PyBBHB

---

PR #61, opened 2026-04-04
refactor: remove unnecessary &mut self from data client, add cache cleanup

DominicBurkart:
$(cat <<'EOF'
### 1. Remove unnecessary `&mut self` from `VelibDataClient`

`InMemoryCache<K, V>` wraps its storage in `Arc<RwLock<HashMap<...>>>`, giving it interior mutability. This means the four public data-fetching methods on `VelibDataClient` — `fetch_reference_stations`, `fetch_realtime_status`, `get_all_stations`, and `get_station_by_code` — never actually mutate the struct through the `self` reference; all mutations go through the lock inside the cache. Declaring them `&mut self` was therefore a false constraint.

**Before:** callers (including `McpToolHandler`) had to acquire an *exclusive write lock* on `Arc<RwLock<VelibDataClient>>` even for pure cache reads.

**After:** the methods take `&self`, so callers can hold a *shared read lock* instead. With `tokio::sync::RwLock` this means multiple concurrent requests can read (and serve) cached station data simultaneously without blocking each other, rather than serialising behind a single writer lock.

All nine `self.data_client.write().await` / `let mut data_client` sites in `McpToolHandler` are updated to `self.data_client.read().await` / `let data_client`.

### 2. Schedule automatic cache cleanup

`InMemoryCache::cleanup_expired()` existed but was never called automatically, meaning entries that have passed their TTL accumulate indefinitely in the `HashMap`. In a long-running server this leads to unbounded memory growth proportional to the number of distinct cache keys inserted over time.

`McpToolHandler::new()` now spawns a background `tokio` task that calls `cleanup_cache()` every 5 minutes. The task holds only a `Weak` reference to the `Arc<RwLock<VelibDataClient>>`, so it does not artificially extend the handler's lifetime: when the last strong `Arc` is dropped the weak upgrade returns `None` and the task exits cleanly.

## Test plan

- [x] `cargo check` passes with zero errors
- [x] Full `cargo test` suite passes (44 tests, 8 ignored as network-dependent)
- [x] Pre-push hook re-runs the full suite and confirms green

EOF
)

### DIFF
--- a/src/data/client.rs
+++ b/src/data/client.rs
@@ -67,7 +67,7 @@ impl VelibDataClient {
     }
 
     /// Fetch all station reference data
-    pub async fn fetch_reference_stations(&mut self) -> Result<Vec<StationReference>> {
+    pub async fn fetch_reference_stations(&self) -> Result<Vec<StationReference>> {
         const CACHE_KEY: &str = "all_reference_stations";
 
         // Check cache first
@@ -125,7 +125,7 @@ impl VelibDataClient {
     }
 
     /// Fetch real-time station status data
-    pub async fn fetch_realtime_status(&mut self) -> Result<HashMap<String, RealTimeStatus>> {
+    pub async fn fetch_realtime_status(&self) -> Result<HashMap<String, RealTimeStatus>> {
         const CACHE_KEY: &str = "all_realtime_status";
 
         // Check cache first
@@ -183,7 +183,7 @@ impl VelibDataClient {
     }
 
     /// Get all stations with optional real-time data
-    pub async fn get_all_stations(&mut self, include_realtime: bool) -> Result<Vec<VelibStation>> {
+    pub async fn get_all_stations(&self, include_realtime: bool) -> Result<Vec<VelibStation>> {
         let reference_stations = self.fetch_reference_stations().await?;
 
         if !include_realtime {
@@ -211,7 +211,7 @@ impl VelibDataClient {
 
     /// Get a specific station by code
     pub async fn get_station_by_code(
-        &mut self,
+        &self,
         station_code: &str,
         include_realtime: bool,
     ) -> Result<Option<VelibStation>> {
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -1,3 +1,5 @@
+use std::time::Duration as StdDuration;
+
 use unicode_normalization::UnicodeNormalization;
 
 use crate::data::VelibDataClient;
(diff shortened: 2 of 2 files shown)


DominicBurkart (commented):
$(cat <<'EOF'
## Code Review

### Critical: Duplicate work with PR #60

**This PR is a strict subset of PR #60 (`test/cache-ttl-handler-proptest`).** Both PRs contain identical changes to `src/mcp/handlers.rs` — the blob SHA for that file is `24b2640a05524ef983fb3bd737bf97117fe3253d` on both branches. PR #60's commit message even explicitly says "Bonus: concurrent-read refactor" and describes the same `&mut self → &self`, `write() → read()`, and `Weak<>`-based cleanup task changes.

These two PRs are on different branches and both target `main`, so they are not stacking — they are duplicating each other. Before either is merged, one of the following should happen:

- Close PR #61 and let PR #60 carry the refactor (since PR #60 also includes the test coverage that validates this behaviour — notably `test_cache_cleanup_removes_expired_entries` and `test_cache_value_expires_after_ttl`).
- Or, if PR #61 is to be merged first, rebase PR #60 on top of it so the cleanup task and `&self` changes aren't committed twice.

Merging both as-is will produce a conflict or a redundant second commit.

---

### Correctness of the `Weak<>` background cleanup task

The pattern is structurally sound. A few observations:

**Stale `Weak` check after upgrade.** In the cleanup loop:

```rust
match weak.upgrade() {
    Some(client) => {
        let guard = client.read().await;
        guard.cleanup_cache().await;
    }
    None => break,
}
```

After `upgrade()` succeeds and `client` is held, the `Arc` strong count is guaranteed to stay above zero for the duration of `cleanup_cache()`. There is no race where the handler could be dropped mid-cleanup and cause undefined behaviour. The `Weak` check is correct.

**No race between cleanup and active reads.** `cleanup_expired()` on `InMemoryCache` takes a `write()` lock on the inner `HashMap`. Normal reads take a `read()` lock. These are mutually exclusive, so cleanup cannot remove an entry while a read is in the middle of using it. However, there is a minor TOCTOU window: a reader could observe a cache hit, find the entry non-expired, return from `get()`, and then have the caller use data that `cleanup_expired()` would have removed a moment later — but since the data was already cloned and returned, there is no problem. The cache data itself is not invalidated retroactively.

**Outer `RwLock<VelibDataClient>` is held as a read lock during cleanup.** The background task does `client.read().await` on the outer `Arc<RwLock<VelibDataClient>>` and then calls `data_client.cleanup_cache()`. Since `VelibDataClient::cleanup_cache()` takes `&self` (interior mutability via the inner `RwLock` on each `InMemoryCache`), the outer read lock does not block other concurrent reads. This is correct and desirable.

**The `interval.tick().await` before the loop skips the immediate first tick.** This means cleanup first runs at T+5 min, not T+0. For a cache with a 2-minute realtime TTL and a 5-minute reference TTL, entries could accumulate for up to 10 minutes (two full intervals) before the first cleanup. This is acceptable for a long-running server, but worth documenting explicitly if memory footprint is a concern.

---

### Does `write() → read()` actually improve throughput?

Yes, under the right conditions. `tokio::sync::RwLock` allows multiple concurrent readers. The previous `write()` call meant every MCP tool invocation — even pure cache reads — serialised behind an exclusive lock. With `read()`, concurrent requests for cached data can proceed in parallel.

The practical benefit depends on request concurrency. In a low-concurrency MCP deployment (single AI agent issuing sequential tool calls), the change has no observable throughput impact. In a higher-concurrency scenario (multiple agents sharing one server instance), the benefit is real.

One subtlety: `VelibDataClient` methods still take `write()` locks internally on `InMemoryCache.entries` when inserting or cleaning up. A cache miss causes a network fetch followed by `insert()` which acquires an inner write lock. Two concurrent cache-miss requests for the same key will both issue network fetches and both write the same data — there is no cache stampede protection (single-flight). This pre-existed this PR, but the switch to outer read locks makes it slightly more likely that two concurrent misses overlap. This is not a regression introduced here, just worth noting.

---

### 5-minute cleanup interval

Reasonable for the cache TTLs in use (2 min realtime, 5 min reference). Expired entries will accumulate for at most one interval (5 min) beyond their TTL before being evicted. For a server with a bounded set of cache keys (effectively two: `"all_reference_stations"` and `"all_realtime_status"`), unbounded growth is not actually a concern regardless of cleanup frequency — but the task is still good hygiene and correctly handles the edge case where key diversity increases in the future.

---

### Summary

The code changes in this PR are correct. The `Weak<>`-based cleanup task is properly structured, the lock downgrade from `write()` to `read()` is safe and beneficial, and the cleanup timing is appropriate. The main issue to resolve before merging is the **duplication with PR #60** — the two PRs should be reconciled rather than both merged independently.

https://claude.ai/code/session_015pVaz1D5CHcLYe9y1SjCyX
EOF
)

---

PR #73, opened 2026-04-05
test: add InMemoryCache TTL tests and JSON parsing unit tests

DominicBurkart:
## Summary

- Adds a `#[cfg(test)] mod tests` block to `src/data/cache.rs` with 5 async (`#[tokio::test]`) tests covering `insert`/`get`, TTL expiry, `cleanup_expired`, `clear`, and `size`.
- Adds a `#[cfg(test)] mod tests` block to `src/data/client.rs` with 5 unit tests covering `parse_realtime_status` (Open / Maintenance / Closed) and `parse_reference_station` (happy path and missing-field error).

## Test plan

- [ ] `cargo test` passes with all 10 new tests green
- [ ] `expired_entry_returns_none` and `cleanup_removes_expired` use `insert_with_ttl(…, Duration::milliseconds(1))` + `tokio::time::sleep(5ms)` — confirm timing is reliable in CI
- [ ] `parse_realtime_status_maintenance` asserts `StationStatus::Maintenance` when `is_installed = "OUI"` but `is_renting = "NON"` — matches actual branch logic in `parse_realtime_status`
- [ ] `parse_reference_station_missing_field_returns_error` asserts `Err` on missing `capacity` — confirms error propagation path

### DIFF
--- a/src/data/cache.rs
+++ b/src/data/cache.rs
@@ -84,3 +84,74 @@ where
         entries.clear();
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::time::Duration as StdDuration;
+
+    #[tokio::test]
+    async fn insert_and_get_returns_value() {
+        let cache: InMemoryCache<String, String> =
+            InMemoryCache::new(Duration::minutes(5));
+        cache.insert("key1".to_string(), "hello".to_string()).await;
+        let result = cache.get(&"key1".to_string()).await;
+        assert_eq!(result, Some("hello".to_string()));
+    }
+
+    #[tokio::test]
+    async fn expired_entry_returns_none() {
+        let cache: InMemoryCache<String, String> =
+            InMemoryCache::new(Duration::minutes(5));
+        // Insert with a 1ms TTL
+        cache
+            .insert_with_ttl(
+                "key_exp".to_string(),
+                "value".to_string(),
+                Duration::milliseconds(1),
+            )
+            .await;
+        // Wait long enough for expiry
+        tokio::time::sleep(StdDuration::from_millis(5)).await;
+        let result = cache.get(&"key_exp".to_string()).await;
+        assert_eq!(result, None);
+    }
+
+    #[tokio::test]
+    async fn cleanup_removes_expired() {
+        let cache: InMemoryCache<String, String> =
+            InMemoryCache::new(Duration::minutes(5));
+        cache
+            .insert_with_ttl(
+                "stale".to_string(),
+                "data".to_string(),
+                Duration::milliseconds(1),
+            )
+            .await;
+        tokio::time::sleep(StdDuration::from_millis(5)).await;
+        cache.cleanup_expired().await;
+        assert_eq!(cache.size().await, 0);
+    }
+
+    #[tokio::test]
+    async fn clear_empties_cache() {
+        let cache: InMemoryCache<String, i32> =
+            InMemoryCache::new(Duration::minutes(5));
+        cache.insert("a".to_string(), 1).await;
+        cache.insert("b".to_string(), 2).await;
+        cache.insert("c".to_string(), 3).await;
+        assert_eq!(cache.size().await, 3);
+        cache.clear().await;
+        assert_eq!(cache.size().await, 0);
+    }
+
+    #[tokio::test]
+    async fn size_reflects_inserts() {
+        let cache: InMemoryCache<u32, &str> =
+            InMemoryCache::new(Duration::minutes(5));
+        for i in 0..7u32 {
+            cache.insert(i, "v").await;
+        }
+        assert_eq!(cache.size().await, 7);
+    }
+}
--- a/src/data/cache.rs
+++ b/src/data/cache.rs
@@ -88,72 +88,149 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::time::Duration as StdDuration;
 
-    #[tokio::test]
-    async fn insert_and_get_returns_value() {
-        let cache: InMemoryCache<String, String> = InMemoryCache::new(Duration::minutes(5));
-        cache.insert("key1".to_string(), "hello".to_string()).await;
-        let result = cache.get(&"key1".to_string()).await;
-        assert_eq!(result, Some("hello".to_string()));
+    #[test]
+    fn cache_entry_not_expired_within_ttl() {
+        let entry = CacheEntry::new(42, Duration::seconds(60));
+        assert!(!entry.is_expired());
+    }
+
+    #[test]
+    fn cache_entry_expired_with_negative_ttl() {
+        let entry = CacheEntry::new(42, Duration::seconds(-1));
+        assert!(entry.is_expired());
     }
 
     #[tokio::test]
-    async fn expired_entry_returns_none() {
-        let cache: InMemoryCache<String, String> = InMemoryCache::new(Duration::minutes(5));
-        // Insert with a 1ms TTL. InMemoryCache uses chrono::Utc::now() for
-        // expiry tracking (not tokio::time::Instant), so tokio::time::pause
-        // cannot mock it. Use a conservative 100ms wall-clock sleep instead
-        // to stay reliable on loaded CI runners.
-        cache
-            .insert_with_ttl(
-                "key_exp".to_string(),
-                "value".to_string(),
-                Duration::milliseconds(1),
-            )
-            .await;
-        tokio::time::sleep(StdDuration::from_millis(100)).await;
-        let result = cache.get(&"key_exp".to_string()).await;
+    async fn insert_and_get_returns_value() {
+        let cache: InMemoryCache<String, i32> = InMemoryCache::new(Duration::seconds(60));
+        cache.insert("key1".to_string(), 100).await;
+
+        let result = cache.get(&"key1".to_string()).await;
+        assert_eq!(result, Some(100));
+    }
+
+    #[tokio::test]
+    async fn get_missing_key_returns_none() {
+        let cache: InMemoryCache<String, i32> = InMemoryCache::new(Duration::seconds(60));
+        let result = cache.get(&"nonexistent".to_string()).await;
         assert_eq!(result, None);
     }
 
     #[tokio::test]
-    async fn cleanup_removes_expired() {
-        let cache: InMemoryCache<String, String> = InMemoryCache::new(Duration::minutes(5));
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
## Review: test/cache-and-parsing

Good addition of unit tests for two core modules. The coverage of the happy path, TTL expiry, and JSON parsing errors is a solid foundation. A few issues worth addressing before merging — one is a likely CI flakiness source, and the rest are gaps that could allow real regressions to slip through.

---

### 1. Timing-based TTL tests are fragile in CI (medium severity)

`expired_entry_returns_none` and `cleanup_removes_expired` both use a 1 ms TTL and a 5 ms `tokio::time::sleep`. On a loaded CI runner these sleeps are not guaranteed — OS scheduling jitter or a slow test executor can easily mean less than 1 ms of real time elapsed inside the sleep, causing the entry to be read before it actually expires. The PR's own test plan flags this as a concern but doesn't resolve it.

The idiomatic Tokio solution is to pause the clock and advance it deterministically:

```rust
#[tokio::test(start_paused = true)]
async fn expired_entry_returns_none() {
    let cache: InMemoryCache<String, String> =
        InMemoryCache::new(Duration::minutes(5));
    cache
        .insert_with_ttl(
            "key_exp".to_string(),
            "value".to_string(),
            Duration::milliseconds(1),
        )
        .await;
    tokio::time::advance(StdDuration::from_millis(10)).await;
    let result = cache.get(&"key_exp".to_string()).await;
    assert_eq!(result, None);
}
```

`start_paused = true` on the `#[tokio::test]` macro freezes `Instant::now()` / `tokio::time` until you explicitly advance it, making the test deterministic and instant. Note that `InMemoryCache` uses `chrono::Utc::now()` rather than `tokio::time::Instant`, so `tokio::time::pause` alone won't mock it — either switch expiry tracking to `tokio::time::Instant` (which _is_ controlled by the paused clock), or accept the wall-clock approach and use a longer, more conservative sleep (e.g. 50–100 ms) to reduce flakiness without adding a new dependency.

---

### 2. `size_reflects_inserts` doesn't test duplicate-key behaviour

```rust
// src/data/cache.rs  –  size_reflects_inserts
for i in 0..7u32 {
    cache.insert(i, "v").await;
}
assert_eq!(cache.size().await, 7);
```

All keys are distinct, so this only exercises the insert-new-key path. A `HashMap` will keep size at 1 if the same key is inserted twice, but if `InMemoryCache` ever wraps a different data structure a re-insert could double-count. Add a second assertion:

```rust
cache.insert(0u32, "updated").await;
assert_eq!(cache.size().await, 7); // duplicate key must not increase size
```

---

### 3. Missing symmetry test for `Maintenance` status

```rust
// src/data/client.rs  –  parse_realtime_status_maintenance
"is_renting": "NON",
"is_returning": "OUI",
```

This covers only one of the two sub-cases that should yield `Maintenance`. The production condition is `if is_renting && is_returning { Open } else { Maintenance }`, so a station that is renting but _not_ returning should also be `Maintenance`. That path is untested. A future author could accidentally change `&&` to `||` and this test would still pass. Add a mirrored case:

```rust
#[test]
fn parse_realtime_status_renting_but_not_returning_is_maintenance() {
    let client = make_client();
    let record = json!({
        "stationcode": "16110",
        "mechanical": 1,
        "ebike": 0,
        "numdocksavailable": 5,
        "is_installed": "OUI",
        "is_renting": "OUI",
        "is_returning": "NON",
        "duedate": "2024-01-15T10:00:00+00:00"
    });
    let (_, status) = client.parse_realtime_status(&record).unwrap();
    assert_eq!(status.status, StationStatus::Maintenance);
}
```

---

### 4. No test for missing `stationcode` in `parse_realtime_status`

`parse_reference_station_missing_field_returns_error` correctly tests that missing `capacity` → `Err`. The analogue for `parse_realtime_status` is missing: there is no test verifying that an absent `stationcode` returns `Err`. This matters because the production ingestion loop silently drops `Err` results — a bug in the early-exit error path would cause stations to vanish with no signal.

```rust
#[test]
fn parse_realtime_status_missing_stationcode_returns_error() {
    let client = make_client();
    let record = json!({
        "mechanical": 2,
        "ebike": 1,
        "numdocksavailable": 8,
        "is_installed": "OUI",
        "is_renting": "OUI",
        "is_returning": "OUI",
        "duedate": "2024-01-15T10:00:00+00:00"
    });
    assert!(client.parse_realtime_status(&record).is_err());
}
```

---

### 5. No test for missing `coordonnees_geo` in `parse_reference_station`

`parse_reference_station_missing_field_returns_error` only covers missing `capacity`. The geo-coordinates block has its own `ok_or_else` error path which goes untested. A regression there would allow stations with no location to silently error out in production.

```rust
#[test]
fn parse_reference_station_missing_coordinates_returns_error() {
    let client = make_client();
    let record = json!({
        "stationcode": "10042",
        "name": "Some Station",
        "capacity": 35
        // coordonnees_geo omitted
    });
    assert!(client.parse_reference_station(&record).is_err());
}
```

---

### Minor nit: `make_client()` creates a real HTTP client

`make_client()` calls `VelibDataClient::new()`, which initialises a live `RetryableHttpClient`. For these unit tests the network is never called, so it works — but a comment would make the intent clearer to future readers, and ideally a `new_for_testing()` constructor (or making the HTTP client injectable) would be the right long-term direction if the project grows to need response mocking.


---

DominicBurkart (commented):
## Follow-up review after rustfmt commit (906559f)

The second commit is a pure formatting fix (collapsing multi-line cache variable declarations). No logic changed, so the substance of the [previous review](https://github.com/DominicBurkart/velib-mcp/pull/73#pullrequestreview-4059970400) still stands in full. This review re-summarises the outstanding issues, adds a new **merge-conflict / redundancy** finding introduced by PR #60 (which post-dates the first review), and corrects one point about the `start_paused` suggestion.

---

### 1. Direct merge conflict with PR #60 — cache.rs (blocker)

PR #60 (`test/cache-ttl-handler-proptest`) also adds a `#[cfg(test)] mod tests` block to `src/data/cache.rs` at the same insertion point (after the `clear` method). Both PRs branch from the same base commit (`21e2dde`). Merging either one first will leave the other in a `dirty` / conflicted state; the conflict cannot be resolved automatically.

**Comparison of the two sets of cache tests:**

| Invariant | PR #73 | PR #60 |
|---|---|---|
| Insert + retrieve | `insert_and_get_returns_value` | `test_cache_insert_and_retrieve` |
| Cache miss (never-inserted key) | — | `test_cache_miss_for_missing_key` |
| TTL expiry | `expired_entry_returns_none` (1 ms TTL, 5 ms sleep) | `test_cache_value_expires_after_ttl` (100 ms TTL, 150 ms sleep) |
| Cleanup removes expired, keeps fresh | `cleanup_removes_expired` (only expired, no fresh entry) | `test_cache_cleanup_removes_expired_entries` (expired + fresh) |
| Clear empties cache | `clear_empties_cache` | — |
| Size count | `size_reflects_inserts` | — |
| Overwrite same key | — | `test_cache_overwrite_existing_key` |
| Remove + verify gone | — | `test_cache_remove` |

PR #60's TTL timing is more conservative (100 ms / 150 ms vs 1 ms / 5 ms), and it tests two additional invariants (`cache_miss` and `cache_overwrite`). PR #73 covers `clear` and `size` which PR #60 omits. The ideal resolution is to merge the two sets rather than discard one — but that requires coordination between the PRs.

**Recommendation:** either rebase this PR on top of PR #60 (or vice versa) and produce a single combined test module, or close one PR and cherry-pick the unique tests into the survivor.

---

### 2. Timing-dependent TTL tests — corrected guidance

The previous review suggested `#[tokio::test(start_paused = true)]`. That suggestion is **not usable here without additional changes**: `InMemoryCache` tracks expiry via `chrono::Utc::now()`, not `tokio::time::Instant`. The paused-clock feature mocks only `tokio::time`; it has no effect on `chrono::Utc::now()`, so the cache entry would still expire based on wall-clock time and the test would remain racy.

The two workable approaches are:

**Option A — increase the sleep margin** (minimal change, still wall-clock based):
```rust
// Use a 50 ms TTL and a 200 ms sleep instead of 1 ms / 5 ms.
// The 4× margin is enough to absorb CI scheduling jitter.
cache.insert_with_ttl("key_exp".to_string(), "value".to_string(),
    Duration::milliseconds(50)).await;
tokio::time::sleep(StdDuration::from_millis(200)).await;
```
PR #60 already uses this pattern (100 ms TTL / 150 ms sleep). Adopting the same margins would make this PR consistent with it.

**Option B — make time injectable** (more robust, more invasive):
Replace `Utc::now()` calls in `CacheEntry` with a clock abstraction so tests can inject a fake clock without sleeping at all. This is the correct long-term direction but out of scope for a test-only PR.

---

### 3. `parse_realtime_status_maintenance` misses one sub-case (from previous review, still open)

The test sets `is_renting = "NON"` and `is_returning = "OUI"`. The production logic is effectively `if is_renting && is_returning { Open } else if is_installed { Maintenance } else { Closed }`. The symmetric case — `is_renting = "OUI"` and `is_returning = "NON"` — is not covered. A future author who changes `&&` to `||` in the Open condition would not be caught by these tests.

---

### 4. `size_reflects_inserts` does not test duplicate-key behaviour (from previous review, still open)

All 7 keys are distinct integers, so the test only exercises the "insert new key" path. A single additional assertion would close the gap:

```rust
cache.insert(0u32, "updated").await;
assert_eq!(cache.size().await, 7); // duplicate must not increase size
```

---

### 5. Missing error-path tests (from previous review, still open)

- No test for `parse_realtime_status` when `stationcode` is absent.
- No test for `parse_reference_station` when `coordonnees_geo` is absent (only `capacity` is tested).

Both of these error paths route through `ok_or_else` in production; if either were accidentally replaced with an `unwrap`, the existing tests would not detect it.

---

### 6. Tokio in dev-dependencies — no issue

`tokio` is declared under `[dependencies]` (not `[dev-dependencies]`) with `features = ["full"]`, so `#[tokio::test]` resolves correctly. No action needed here.

---

### Summary

The rustfmt commit is clean. The substantive blocker introduced since the last review is the **direct structural conflict with PR #60**, which adds an overlapping `mod tests` block to the same file at the same location. The timing fragility in the TTL tests (issue 1 from the previous review) is real but the fix is simpler than suggested: increase the sleep margin to match PR #60's 100 ms / 150 ms pattern rather than switching to a paused clock. Issues 3–5 from the previous review remain open.

---

PR #74, opened 2026-04-05
arch: extract retry execution helper and remove dead MCP types

DominicBurkart:
### Issue 1 — Deduplicate retry execution logic (`src/data/retry.rs`)

`get()` and `get_with_query()` previously each contained an identical ~20-line closure passed to `retry_policy.execute()`: `send()` call, debug log, 429 detection with `warn!`, rate-limit error creation, non-success status check. The only difference was whether `.query(query)` was chained before `.send()`.

A new private `execute_with_retry(&self, url, build_request)` method now owns all of that logic. It accepts a `Fn() -> RequestBuilder` closure so the caller controls how the builder is constructed (with or without query params). Both public methods become one-liners that build the `RequestBuilder` and call through to it.

### Issue 2 — Remove dead types + replace glob re-exports

**Dead type removal (`src/mcp/types.rs`):**  
Confirmed via `mcp__github__search_code` that `StationListResponse`, `ResponseMetadata`, `PaginationInfo`, `GeographicQuery`, and `StationQuery` appear only in their own definitions and a doc markdown file — never referenced in any Rust source. All five structs (and the `default_limit` free function that existed solely to supply `GeographicQuery`'s field default) have been removed.

**Explicit re-exports (`src/lib.rs`, `src/mcp/mod.rs`):**  
- `src/lib.rs`: `pub use types::*` → explicit list of the 10 types from `src/types.rs` that actually exist
- `src/mcp/mod.rs`: `pub use types::*` → explicit list of the 23 types from `src/mcp/types.rs` that are referenced by `handlers.rs` and `server.rs`

## Test plan

- [ ] `cargo check` passes with no warnings about unused imports or dead code
- [ ] `cargo test` passes — all existing retry policy unit tests and MCP types tests still green
- [ ] Confirm `get()` and `get_with_query()` behave identically to before (same retry/error path, only request builder differs)
- [ ] Confirm the 5 removed types are not imported anywhere else in the codebase

PR #75, opened 2026-04-05
arch: remove dead code, fix fabricated metric, deduplicate test helpers

DominicBurkart:
## Summary

Four focused architecture improvements to reduce complexity and eliminate misleading code:

### 1. Extract shared `find_available_port` into `tests/common/mod.rs`

`tests/smoke_test.rs` and `tests/integration_test.rs` both defined an identical `find_available_port()` function. The function is now in `tests/common/mod.rs` and both test files use `mod common; use common::find_available_port;` instead. Eliminates copy-paste drift risk.

### 2. Remove dead `clients: Arc<RwLock<HashMap<String, WebSocketClient>>>` from `McpServer`

The `clients` map was populated on WebSocket connect/disconnect but never read anywhere in the codebase. The `WebSocketClient` struct existed solely to be stored in this map and carried `#[allow(dead_code)]` on its only field (`id: String`). Both the field and the struct are removed, along with the `Arc::clone(&clients)` threading through the router and the insert/remove calls in the WebSocket handler.

### 3. Fix the fabricated cache hit-rate formula in `get_health_resource`

The previous `hit_rate` value was computed as `0.75 + (total_entries as f64 / 2000.0) * 0.2` — a made-up number that had no relationship to actual cache hits or misses. It was replaced with a plain `entry_count` field reporting the real total number of cached entries. The misleading `hit_rate` key is removed from the response entirely.

### 4. Remove redundant `is_valid_paris_metro` bounding-box check

The Paris metro bounding box (roughly 48.7–49.0 N, 2.0–2.6 E) is entirely contained within the 50 km `is_within_paris_service_area` radius centred on Paris City Hall. The `is_valid_paris_metro()` guard was therefore a strict subset of `is_within_paris_service_area()` and could never reject a coordinate that the latter would accept. The redundant calls are removed from `find_nearby_stations` and `plan_bike_journey`, leaving only the radius check.

## Test plan

- [ ] `cargo build` compiles cleanly with no dead-code warnings for the removed items
- [ ] `cargo test` passes (smoke and integration tests resolve `common::find_available_port` correctly)
- [ ] `/health` and `velib://health` resource responses no longer contain a `hit_rate` field; `entry_count` is present instead
- [ ] Coordinates inside the service area but outside the old bbox (e.g. banlieue just beyond the bbox corners) are now accepted by `find_nearby_stations` and `plan_bike_journey`

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -70,12 +70,6 @@ impl McpToolHandler {
         }
 
         let query_point = Coordinates::new(input.latitude, input.longitude);
-        if !query_point.is_valid_paris_metro() {
-            return Err(Error::InvalidCoordinates {
-                latitude: input.latitude,
-                longitude: input.longitude,
-            });
-        }
 
         // Enforce 50km distance limit from Paris City Hall
         if !query_point.is_within_paris_service_area() {
@@ -279,20 +273,6 @@ impl McpToolHandler {
         &self,
         input: PlanBikeJourneyInput,
     ) -> Result<PlanBikeJourneyOutput> {
-        if !input.origin.is_valid_paris_metro() {
-            return Err(Error::InvalidCoordinates {
-                latitude: input.origin.latitude,
-                longitude: input.origin.longitude,
-            });
-        }
-
-        if !input.destination.is_valid_paris_metro() {
-            return Err(Error::InvalidCoordinates {
-                latitude: input.destination.latitude,
-                longitude: input.destination.longitude,
-            });
-        }
-
         // Enforce 50km distance limit from Paris City Hall for both origin and destination
         if !input.origin.is_within_paris_service_area() {
             let distance_km = input.origin.distance_to(&PARIS_CITY_HALL) / 1000.0;
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -357,7 +357,8 @@ impl McpToolHandler {
             let pickup_walk_ratio = f64::from(best_pickup.straight_line_distance_meters) / max_walk;
             let dropoff_walk_ratio =
                 f64::from(best_dropoff.straight_line_distance_meters) / max_walk;
-            let confidence_score = 1.0 - f64::midpoint(pickup_walk_ratio, dropoff_walk_ratio) * 0.5;
+            let confidence_score =
+                1.0 - f64::midpoint(pickup_walk_ratio, dropoff_walk_ratio) * 0.5;
 
             recommendations.push(JourneyRecommendation {
                 pickup_station: best_pickup.station.clone(),
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -7,10 +7,8 @@ use axum::{
 };
 use chrono::Utc;
 use serde_json::{json, Value};
-use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
-use tokio::sync::RwLock;
 use tracing::{error, info, warn};
 
 use super::handlers::McpToolHandler;
@@ -19,15 +17,7 @@ use crate::{Error, Result};
 
 pub struct McpServer {
     tool_handler: Arc<McpToolHandler>,
-    clients: Arc<RwLock<HashMap<String, WebSocketClient>>>,
     start_time: Instant,
-}
-
-#[derive(Debug)]
-struct WebSocketClient {
-    #[allow(dead_code)]
-    id: String,
-    // Additional client metadata can be added here
 }
 
 impl Default for McpServer {
@@ -41,14 +31,12 @@ impl McpServer {
     pub fn new() -> Self {
         Self {
             tool_handler: Arc::new(McpToolHandler::new()),
-            clients: Arc::new(RwLock::new(HashMap::new())),
             start_time: Instant::now(),
         }
     }
 
     pub fn router(&self) -> Router {
         let handler = Arc::clone(&self.tool_handler);
-        let clients = Arc::clone(&self.clients);
 
         Router::new()
             .route(
@@ -90,10 +78,9 @@ impl McpServer {
                 "/mcp/ws",
                 get({
                     let handler = Arc::clone(&handler);
-                    let clients = Arc::clone(&clients);
                     move |ws: WebSocketUpgrade| async move {
                         ws.on_upgrade(move |socket| {
-                            Self::handle_websocket_connection(socket, handler, clients)
+                            Self::handle_websocket_connection(socket, handler)
                         })
                     }
                 }),
@@ -114,21 +101,9 @@ impl McpServer {
     async fn handle_websocket_connection(
         mut socket: WebSocket,
         handler: Arc<McpToolHandler>,
-        clients: Arc<RwLock<HashMap<String, WebSocketClient>>>,
     ) {
         let client_id = uuid::Uuid::new_v4().to_string();
         info!("New WebSocket connection: {}", client_id);
-
-        // Add client to the map
-        {
-            let mut clients_guard = clients.write().await;
-            clients_guard.insert(
-                client_id.clone(),
-                WebSocketClient {
-                    id: client_id.clone(),
-                },
-            );
-        }
 
         // Handle messages
         while let Some(msg) = socket.recv().await {
@@ -205,12 +180,6 @@ impl McpServer {
                 }
                 _ => {} // Ignore other message types
             }
-        }
-
-        // Remove client from the map
-        {
-            let mut clients_guard = clients.write().await;
-            clients_guard.remove(&client_id);
         }
 
         info!("WebSocket connection terminated: {}", client_id);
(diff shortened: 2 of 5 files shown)


DominicBurkart (commented):
## Code Review — `arch/reduce-complexity`

Overall this is a clean, well-scoped refactor. The four changes are coherent, well-described, and directionally correct. I have one **bug** (inconsistency left behind by change 4), one **moderate concern** (data-contract break in the health resource), and a couple of minor observations.

---

### Bug — `is_valid_paris_metro` is still called inside `StationReference::validate()` after being removed from the handlers

**File:** `src/types.rs`, inside `StationReference::validate()`

```rust
if !self.coordinates.is_valid_paris_metro() {
    return Err("Coordinates are outside valid Paris metro area".to_string());
}
```

The PR removes `is_valid_paris_metro()` guards from `handlers.rs` (change 4) but leaves this call intact in `StationReference::validate()`. The function itself also remains in `types.rs`. This means:

1. The stated goal — "leaving only the radius check" — is not fully achieved. Station data ingested from the open-data API could still be silently rejected at the `validate()` level if a legitimate station sits inside the 50 km radius but outside the narrow bounding box (e.g. a station in a banlieue commune whose coordinates fall just outside `lon >= 2.0 && lon <= 2.6`).
2. `is_valid_paris_metro` is now dead code at the API-validation layer but live code at the data-model layer, creating a confusing split personality for the function.

The fix is either:
- Replace the `is_valid_paris_metro()` call in `StationReference::validate()` with `is_within_paris_service_area()` (consistent with the handler change), or
- Remove `is_valid_paris_metro()` entirely if the bbox is no longer wanted anywhere.

The test `test_coordinates_paris_validation` in `types.rs` also only exercises `is_valid_paris_metro`; if the function is removed that test needs updating.

---

### Moderate concern — breaking change to `/health` and `velib://health` response shape

**File:** `src/mcp/server.rs`, `get_health_resource`

```diff
-"hit_rate": hit_rate.min(1.0),
-"entries": total_entries,
+"entry_count": total_entries,
```

Two keys are removed (`hit_rate`, `entries`) and one new key is added (`entry_count`). Any client, dashboard, or integration test that reads `cache_stats.entries` or `cache_stats.hit_rate` will silently break. The fix to the fabricated formula is correct and welcome, but the rename from `entries` to `entry_count` is a gratuitous breaking change that could be avoided by keeping the key name as `entries`. If a deliberate API-version bump is intended, that should be called out explicitly (changelog, semver note, or doc update). As it stands the PR description mentions removing `hit_rate` but does not mention renaming `entries`.

---

### Minor — `find_available_port` TOCTOU race

**File:** `tests/common/mod.rs`

The helper binds to port 0 to discover a free port, then drops the listener and returns the port number. The test harness then tries to bind that same port when starting the server. There is a small window between the two binds where another process can claim the port, causing a flaky test. This pre-exists the PR (it was in both test files before), but the extraction into a shared helper is a good moment to fix it — consider returning the bound `TcpListener` itself and having the server code accept a pre-bound listener, or at minimum documenting the known race.

---

### Minor — `WebSocketClient` / `clients` removal leaves no connection-count observability

**File:** `src/mcp/server.rs`

The removal is correct (dead code is dead code), but it is worth noting that once this merges there is no facility to observe how many WebSocket clients are currently connected. If connection-count metrics are ever wanted for the health endpoint or for capacity planning, the pattern to add them back would be an `AtomicUsize` counter rather than a `HashMap` — just a forward-looking note, not a blocker.

---

### Summary table

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **Bug** | `src/types.rs` `StationReference::validate` | `is_valid_paris_metro` still called; contradicts change 4 |
| 2 | Moderate | `src/mcp/server.rs` `get_health_resource` | `entries` key renamed to `entry_count` — undocumented breaking change |
| 3 | Minor | `tests/common/mod.rs` | TOCTOU race in `find_available_port` (pre-existing, good time to address) |
| 4 | Nit | `src/mcp/server.rs` | No connection-count observability after `clients` map removal |

Item 1 should be resolved before merge. Item 2 needs either a decision to keep `entries` or an explicit note that this is an intentional API break.

---

DominicBurkart (commented):
## Code Review

Four changes, three are clean — one has a semantic risk that deserves explicit validation or documentation before merge.

---

### 1. `find_available_port` deduplication — LGTM

`tests/common/mod.rs` declares `pub fn find_available_port()` with no visibility qualifier on the module itself (which is correct for Rust integration-test layout). Both `tests/smoke_test.rs` and `tests/integration_test.rs` use `mod common; use common::find_available_port;` — this is the idiomatic way to share helpers across integration-test files in a Cargo workspace and will compile cleanly. No issues.

---

### 2. Dead `clients: Arc<RwLock<HashMap<String, WebSocketClient>>>` removal — LGTM with one observation

The diff confirms the map was insert-only: nothing in the codebase ever called `.get()` or iterated over it, and `WebSocketClient` carried `#[allow(dead_code)]` on its sole field. Removing the map, the struct, the `Arc::clone` in `router()`, and the insert/remove calls in `handle_websocket_connection` is correct.

One observation: the `client_id` UUID is still generated and logged at connect/disconnect, but is no longer used for anything beyond those two log lines. That is fine for traceability, but if future work needs per-connection state (e.g. authentication, per-client subscriptions) the scaffolding will need to be re-added. A short `// TODO: re-introduce client map if per-connection state is needed` comment near the `client_id` line would make intent explicit — not blocking, but considerate.

---

### 3. Cache hit-rate fix — LGTM

`get_health_resource` now calls `handler.cache_stats().await` which delegates to `VelibDataClient::cache_stats()`. Reading `handlers.rs` confirms `cache_stats()` returns `(reference_cache_size, realtime_cache_size)` from the real in-memory cache, so `total_entries = reference_cache_size + realtime_cache_size` is a genuine live measurement. Replacing the fabricated `0.75 + (total_entries / 2000.0) * 0.2` formula with the plain `entry_count` field is correct.

One note: the JSON key rename from `entries` to `entry_count` is a breaking change to the health resource schema. If any client or dashboard consumes this field by key name it will silently lose the data. Worth a line in release notes.

---

### 4. `is_valid_paris_metro` removal — NEEDS FOLLOW-UP

The PR description argues that the bbox (48.7–49.0 N, 2.0–2.6 E) is "entirely contained within" the 50 km radius and therefore the bbox check is a redundant strict subset. **This claim is geometrically incorrect for at least one bbox corner.**

Paris City Hall: 48.8565 N, 2.3514 E. 50 km radius.

Haversine estimates for the four bbox corners (using the same formula in `src/types.rs`):
- SW corner (48.7 N, 2.0 E): ~30 km — inside the circle ✓
- NW corner (49.0 N, 2.0 E): **~52 km — outside the 50 km circle** ✗
- NE corner (49.0 N, 2.6 E): ~49 km — borderline/inside ✓
- SE corner (48.7 N, 2.6 E): ~24 km — inside the circle ✓

The NW corner of the bbox is approximately 52 km from City Hall — further than 50 km. The two checks are therefore *not* in a strict subset relationship. The `is_within_paris_service_area` radius check would already reject that corner too, so there is no regression in observable rejection behaviour. But the justification in the PR description is wrong, which matters because it is the only documented rationale for the change.

The more impactful (and correct) consequence runs the other direction: coordinates **outside the bbox but inside the 50 km circle** (e.g. 49.1 N, 2.35 E — ~27 km north of City Hall, latitude > 49.0) were previously bbox-rejected and will now correctly pass. That widening is the right policy outcome.

**Two concrete follow-up items:**

**a) Fix the PR description** — the bbox is not a strict subset of the 50 km circle; neither contains the other. State the actual rationale: the radius check is the authoritative policy boundary and the bbox check is being removed as an inconsistent redundant guard.

**b) `StationReference::validate()` in `src/types.rs` still calls `is_valid_paris_metro()`** (around line 155 in the post-merge file). If a station returned by the Paris Open Data API has coordinates that are within 50 km but outside the bbox (legitimately valid), `validate()` will reject it. This is an inconsistency: the handler layer now uses the radius check exclusively, but the data-layer validation still uses the stricter bbox. Either replace `is_valid_paris_metro()` with `is_within_paris_service_area()` in `StationReference::validate()`, or add a comment explaining why the tighter bbox is intentional at the data layer.

---

### Conflict analysis

- **PR #74** (`src/mcp/handlers.rs`): Changes every `write()` lock to `read()` and adds a background cleanup task — significant edits to `handlers.rs`. PR #75 also edits `handlers.rs` (removes the `is_valid_paris_metro` guards). These two PRs touch overlapping hunks; whichever merges second will have a textual conflict. **Coordinate merge order carefully.**

- **PR #60** (`src/types.rs`, `src/data/client.rs`, `Cargo.toml`): Adds proptest-based Haversine property tests to `src/types.rs`. PR #75 does not touch `src/types.rs`, so no direct conflict — but both PRs modify the test suite for `types.rs` in spirit. Verify the test modules do not collide after rebase.

- **PR #55 / #73**: Not retrieved; unable to assess.

---

### Summary

Items that should be addressed before merge:
1. Correct the PR description for change #4 (the bbox-is-a-subset claim is false).
2. Update `StationReference::validate()` in `src/types.rs` to use `is_within_paris_service_area()` for consistency with the handler layer, or document why the bbox is intentionally stricter there.
3. Coordinate with PR #74 — overlapping edits in `handlers.rs` will conflict on merge.


---

PR #96, opened 2026-04-17
test(velib-mcp): improve coverage and test quality

DominicBurkart:
## What was tested

A new offline test file `tests/domain_invariants_tests.rs` (44 tests, zero network calls) targeting the core domain types in `src/types.rs` and `src/mcp/types.rs`.

### Why this file / why these types

Every existing test file either required a live network connection (`data_client_tests.rs`, `integration_test.rs`, `smoke_test.rs`) or focused on handler-level validation that rejected bad inputs *before* touching business logic. The domain types themselves — `DataFreshness`, `BikeAvailability`, `Coordinates`, `StationReference`, `VelibStation`, `GeographicBounds`, `AvailabilityFilter` — had only shallow inline tests inside `src/types.rs` that missed every boundary and fallback path.

## Current disposition

All 44 new tests are pure unit tests; they compile and run with `cargo test --test domain_invariants_tests` on the branch without any environment variables or network access.

## Key invariants now validated

| Area | Invariants |
|---|---|
| `DataFreshness::from_age` | Exact boundary values at **10.0**, **30.0**, **120.0** minutes — the `<` comparisons make these off-by-one traps; tests pin the correct bucket for `x - ε`, `x`, and `x + ε` |
| `BikeAvailability::total` | `saturating_add` must not panic or wrap at `u16::MAX`; both-max case confirmed |
| `Coordinates::distance_to` | Self-distance ≈ 0; symmetry `d(a,b) == d(b,a)`; known Louvre→Notre-Dame distance ≈ 1 km ± 10 %; monotonicity with increasing separation |
| `StationReference::validate` | All five error branches: empty code, empty name, zero capacity, capacity > 200, coordinates outside Paris metro; boundary at exactly 200 (allowed) |
| `VelibStation` no-real-time fallbacks | `is_operational()` returns `true` when `real_time` is `None`; `has_available_bikes` and `has_available_docks` return `false` when `real_time` is `None` |
| `VelibStation::validate` capacity | bikes + docks > capacity → error; bikes + docks == capacity → ok |
| `GeographicBounds::contains` | Degenerate point-bounds; inverted latitude (south > north) contains nothing; inverted longitude contains nothing |
| `AvailabilityFilter` serde | Absent `exclude_out_of_service` field deserializes to `true` (the `default_true` fn); explicit `false` is respected; optional fields absent → `None` |


tests/handler_validation_tests.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[tokio::test]
+async fn find_nearby_rejects_coordinates_outside_service_area() {
+    let handler = McpToolHandler::new();
+    // Coordinates within Paris metro bounds but > 50km from city hall
+    // (far eastern edge of the metro box)
+    let input = FindNearbyStationsInput {
+        latitude: 48.95,
+        longitude: 2.59,
+        radius_meters: 500,
```
DominicBurkart, line 73: Tautological / vacuous test: the branches `Ok(_)`, `outside_service_area`, and `http_error` are all accepted, so this assertion effectively cannot fail for any handler behavior short of a panic. That defeats the purpose of a validation test and will silently hide regressions (e.g., if the handler starts returning `invalid_coordinates` or a different error type here, the test still passes). Either (a) pick inputs that deterministically trigger `OutsideServiceArea` and assert that exact error_type, or (b) delete this test — it pads coverage without validating behavior. Also note: the comment says "if it passes validation, the API call will follow" — this means this test may actually perform a live network request depending on the computed distance, which contradicts the file's offline-first framing.

---

tests/handler_validation_tests.rs
```diff
@@ -0,0 +1,148 @@
...
+    let result = handler.find_nearby_stations(input).await;
+    // This might be within service area depending on exact distance calculation,
+    // but if it passes validation, the API call will follow. We just verify
+    // the validation layer doesn't panic.
+    // The key point is that if the coordinates are truly > 50km away,
+    // we get an OutsideServiceArea error.
+    if let Err(err) = &result {
+        let err_type = err.error_type();
+        assert!(
+            err_type == "outside_service_area" || err_type == "http_error",
```
DominicBurkart, line 87: `search_stations_rejects_short_query` only asserts `result.is_err()`. This is loose: the test passes even if the handler errored for an unrelated reason (e.g. network failure, panic-converted-to-error). Please assert the specific `error_type()` (e.g. `"validation_error"`) so a change in failure mode doesn't slip through unnoticed, matching the style of the neighboring tests.

---

tests/domain_invariants_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+fn coordinates_distance_increases_with_separation() {
+    let origin = Coordinates::new(48.8565, 2.3514);
+    let near = Coordinates::new(48.8575, 2.3514); // ~110 m north
+    let far = Coordinates::new(48.8665, 2.3514); // ~1.1 km north
+    assert!(origin.distance_to(&near) < origin.distance_to(&far));
+}
+
+// ---------------------------------------------------------------------------
+// StationReference::validate – every error branch
+// ---------------------------------------------------------------------------
```
DominicBurkart, line 175: `station_reference_validate_out_of_paris_bounds_is_err` accepts any of three substrings (`"Paris"`, `"coordinate"`, `"Coordinates"`). This is brittle in the wrong direction — it is so permissive that virtually any error message passes. Pin the exact expected message or at minimum a single canonical substring; otherwise a copy-paste error in the production message ("Paris" -> "paris", or a refactor to "outside metro area") will be caught by neither code review nor this test. Same pattern appears in the empty-code / empty-name / zero-capacity tests above — consider asserting on a stable enum/error-code field rather than free-form message text.

---

tests/domain_invariants_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+#[test]
+fn coordinates_distance_is_symmetric() {
+    let a = Coordinates::new(48.8566, 2.3522);
+    let b = Coordinates::new(48.8606, 2.3376);
+    let d_ab = a.distance_to(&b);
+    let d_ba = b.distance_to(&a);
+    // The two values should agree to within floating-point rounding.
+    assert!(
+        (d_ab - d_ba).abs() < 1e-6,
+        "distance not symmetric: {d_ab} vs {d_ba}"
```
DominicBurkart, line 147: The "known reference pair" test uses a 900-1200 m window (roughly +/- 14%) for a pair whose true geodesic distance is ~1.05 km. Haversine on coordinates given to 4 decimal places is accurate to well under 1%, so this tolerance is far looser than needed and will mask real regressions (e.g. a unit bug producing 600 m or 1.5 km). Tighten to, say, 1000-1100 m (+/- 5%) so this is an actual correctness assertion rather than a smoke test.

---

tests/domain_invariants_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+
+#[test]
+fn station_reference_validate_happy_path() {
+    assert!(paris_ref("12345").validate().is_ok());
+}
+
+// ---------------------------------------------------------------------------
+// VelibStation – no-real-time fallback paths
+// ---------------------------------------------------------------------------
+
```
DominicBurkart, line 237: `velib_station_validate_bikes_plus_docks_exceeds_capacity_is_err`: 6 mechanical + 5 electric is 11, and `BikeAvailability::new` may or may not cap that to capacity-aware values depending on the production code. More importantly, this test only checks `is_err()` without checking *why* — if the station also fails for, e.g., negative dock counts or freshness reasons, the test passes for the wrong reason. Please assert the error message mentions capacity / overflow, mirroring the pattern of the `StationReference` tests. Missing edge: test the boundary at bikes+docks == capacity+1 (one over) rather than 16 vs 10, to actually exercise the off-by-one boundary.

---

src/data/cache.rs
```diff
@@ -84,3 +84,141 @@ where
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn cache_entry_not_expired_within_ttl() {
+        let entry = CacheEntry::new(42, Duration::seconds(60));
+        assert!(!entry.is_expired());
+    }
+
```
DominicBurkart, line 97: Using `Duration::seconds(-1)` to simulate expiry is a control-flow shortcut rather than a genuine time-based test. It verifies the `is_expired()` branch fires when `expires_at < now` but does not validate that a genuinely inserted entry ages out after its TTL elapses. Consider using `tokio::time::pause()` + `advance()` (or a mockable `Clock`) to cover the realistic path — "inserted with 60s TTL, advance 61s, now expired". As written, a bug where TTL arithmetic is inverted (e.g. `now - ttl` instead of `now + ttl`) could still pass these tests because negative durations short-circuit the logic.

---

tests/cache_tests.rs
```diff
@@ -0,0 +1,44 @@
...
+            cache.cleanup_expired().await;
+        }));
+    }
+
+    for handle in handles {
+        handle.await.expect("Task should not panic");
+    }
+
+    // All 20 writes should be visible
+    assert_eq!(cache.size().await, 20);
```
DominicBurkart, line 43: This "concurrent access" test only proves the code doesn't deadlock or panic — it asserts nothing about correctness under contention (lost updates, stale reads, cleanup racing with insert). With 20 unique keys and no overlap between writers, there is effectively no contention to exercise. To meaningfully test concurrency, use overlapping keys (e.g. all tasks hammer key `0..4`) and either assert a monotonic invariant or use a tool like `loom` / `shuttle`. Otherwise consider deleting — this reads as coverage padding rather than a behavioral test.

---

src/error.rs
```diff
@@ -82,3 +82,195 @@ impl Error {
...
+                    longitude: 0.0,
+                },
+                -32602,
+            ),
+            (Error::OutsideServiceArea { distance_km: 100.0 }, -32602),
+            (
+                Error::SearchRadiusTooLarge {
+                    radius: 10000,
+                    max: 5000,
+                },
```
DominicBurkart, line 105: These `mcp_error_code_*` tests mirror the match arms one-for-one — the assertion list is structurally identical to the implementation in `mcp_error_code`. This is the textbook tautological test: if someone refactors the code by copy-pasting the wrong constant into both places, both files change together and the test still passes. Consider instead asserting against the JSON-RPC spec directly (cite the constant names: `PARSE_ERROR = -32700`, `INVALID_REQUEST = -32600`, `INVALID_PARAMS = -32602`, `INTERNAL_ERROR = -32603`) via named `const` values imported from a shared module. Same feedback applies to `error_type_covers_all_non_http_variants`.

---

tests/domain_invariants_tests.rs
DominicBurkart, line 1: Missing edges worth adding to this otherwise strong file:

- `Coordinates::new` with NaN/infinity inputs (does `distance_to` panic? produce NaN? propagate?). Without this, a field-order swap or NaN-poisoning bug will not be caught.
- `GeographicBounds::contains` when a point is exactly on the boundary (north/east inclusive vs exclusive) — currently only the degenerate-point case touches the boundary.
- `DataFreshness::from_age` with `f64::NAN` and `f64::INFINITY` — since `<` against NaN is always false, NaN likely falls through to `VeryStale`, which may be wrong for clock-skew recovery.
- `AvailabilityFilter` with all-None + round-trip through serde_json to catch accidental `#[serde(skip_serializing_if)]` removals.

No PII/secret concerns spotted — coordinates used are all public landmarks.

---

DominicBurkart (commented):
Test-quality review for PR #96 ("improve coverage and test quality"). Overall this is a solid, mostly-offline expansion of the suite with good boundary coverage on `DataFreshness`, `StationReference`, and `BikeAvailability`. Flagging several issues below:

**Tautological / coverage-padding patterns**
- `error.rs` unit tests for `mcp_error_code` and `error_type` mirror the match arms one-for-one — they lock in the current mapping but don't validate it against an external spec.
- `tests/cache_tests.rs::cache_concurrent_access_does_not_deadlock` asserts only the absence of deadlock/panic with non-overlapping keys, so it exercises almost no real concurrency.

**Brittle / permissive assertions**
- `station_reference_validate_out_of_paris_bounds_is_err` accepts any of three substrings; tighten to one.
- `search_stations_rejects_short_query` only checks `is_err()` — should assert `error_type()`.
- `find_nearby_rejects_coordinates_outside_service_area` accepts Ok OR two different error types, making it unfalsifiable; it may also hit the network.

**Test-shape issues**
- `src/data/cache.rs` expiry tests use `Duration::seconds(-1)` — a control-flow shortcut that wouldn't catch sign-inverted TTL arithmetic. Use `tokio::time::pause/advance` instead.
- Louvre→Notre-Dame distance tolerance (±14%) is far looser than Haversine warrants; tighten to ±5%.
- `velib_station_validate_bikes_plus_docks_exceeds_capacity_is_err` tests 16 vs 10; better to exercise the capacity+1 boundary and assert on error message.

**Missing edges**: NaN/Inf in `Coordinates` + `DataFreshness::from_age`, exact-boundary inclusivity in `GeographicBounds::contains`, serde round-trip of `AvailabilityFilter`.

No PII or secrets introduced. `.env.deploy.example` is a template, not a real env. Helpers (`tests/common/mod.rs::find_available_port`) look appropriate.

---

PR #101, opened 2026-04-19
[arch] velib-mcp: fix ignored input fields and misclassified validation error

DominicBurkart:
## Summary

Three input fields were silently ignored by their handlers, and a validation failure was misclassified as an internal error:

- **`get_station_by_code`**: `include_real_time` from the input was ignored; the handler always passed `true` to the data client. Now the input field is honoured.
- **`get_area_statistics`**: `include_real_time` from the input was ignored; the handler always fetched real-time data. Now the input field is honoured.
- **`search_stations_by_name`**: a query shorter than 2 characters returned `Error::Internal`, which maps to MCP code `-32603`. This is a client input error, not an internal error — changed to `Error::Validation` (MCP code `-32602`).
- **`find_nearby_stations`**: `AvailabilityFilter.min_bikes` and `AvailabilityFilter.min_docks` were deserialized but never applied. Both predicates are now enforced before returning results.

## Test plan

- [ ] `cargo test` passes (unit tests in `src/mcp/handlers.rs` cover the new `min_bikes` / `min_docks` paths and the `validation_error` error type)
- [ ] Verify `get_station_by_code` with `include_real_time: false` no longer triggers a real-time API call (observable via logs / reduced latency)

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -123,21 +123,41 @@ impl McpToolHandler {
         let mut data_client = self.data_client.write().await;
         let all_stations = data_client.get_all_stations(true).await?;
 
-        // Filter stations by distance and bike type
+        // Filter stations by distance and availability
         let stations = find_stations_within_radius(
             &all_stations,
             &query_point,
             input.radius_meters,
             input.limit as usize,
             |station| {
-                let has_requested_bikes = match &input.availability_filter {
-                    Some(filter) => match &filter.bike_type {
-                        Some(bike_type) => station.has_available_bikes(bike_type),
-                        None => true,
-                    },
-                    None => true,
-                };
-                has_requested_bikes && station.is_operational()
+                if !station.is_operational() {
+                    return false;
+                }
+                if let Some(filter) = &input.availability_filter {
+                    // Apply bike type filter
+                    if let Some(bike_type) = &filter.bike_type {
+                        if !station.has_available_bikes(bike_type) {
+                            return false;
+                        }
+                    }
+                    // Apply minimum bikes filter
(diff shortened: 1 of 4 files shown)


DominicBurkart (commented):
[CRON code review - session 3a0811bb]

Verified against d04d94f.

Plumbing fixes (all correctly wired):
- `get_station_by_code` now forwards `input.include_real_time` instead of hardcoded `true`.
- `get_area_statistics` now passes `input.include_real_time` to `get_all_stations`.
- `find_nearby_stations` applies `min_bikes` (via `rt.bikes.total()`) and `min_docks` (via `has_available_docks`) inside the filter closure; operational check preserved.

Error reclassification: `Error::Validation` maps to `-32602` / `validation_error` in `src/error.rs`, so the short-query path now returns invalid-params rather than internal-error. Correct.

Test coverage: unit tests for min_bikes/min_docks filters and the Validation error type/code; integration test `tools_call_search_short_query_returns_validation_error` asserts both `error_type == "validation_error"` and `code == -32602`. `handler_validation_tests::search_stations_rejects_short_query` updated to assert the new error_type. AvailabilityFilter defaults/round-trip tested.

Minor nits (non-blocking):
- `min_bikes` predicate in handler inlines `rt.bikes.total()` rather than reusing a helper like `has_available_bikes`; slight asymmetry with the docks path but acceptable.
- `test_min_bikes_filter_applied` / `test_min_docks_filter_applied` re-implement the filter inline rather than exercising the real handler closure, so they validate the logic shape but not the production wiring. The jsonrpc integration test partially compensates.
- `get_area_statistics_empty_bounds_does_not_panic` tolerates network errors — fine for this PR's scope.

LGTM.

---

PR #107, opened 2026-04-20
refactor: deduplicate MCP request plumbing and centralize Paris reference point

DominicBurkart:
Automated improvement PR — please review before merging.

## Motivation

While reading `CLAUDE.md`, `README.md`, and the `src/` tree, three concrete alignment / redundancy issues stood out. The architecture docs emphasize "Patterns Rust idiomatiques respectés" and "Séparation responsabilités claire", but a few hotspots carried repeated boilerplate and a duplicated magic constant. Each change below is surgical and behavior-preserving.

### 1. Centralize the Paris City Hall reference point (`src/types.rs`, `src/mcp/handlers.rs`)

- **Before:** `Coordinates::is_within_paris_service_area` hard-coded `PARIS_CITY_HALL_LAT` / `_LON` as function-local `const`s, and `src/mcp/handlers.rs` defined its own `PARIS_CITY_HALL: Coordinates` constant, then inlined `query_point.distance_to(&PARIS_CITY_HALL) / 1000.0` in three separate places to build `OutsideServiceArea` errors.
- **After:** One public `PARIS_CITY_HALL` constant plus `PARIS_SERVICE_AREA_MAX_METERS` live in `src/types.rs`, and `Coordinates::distance_to_paris_city_hall_km()` is the single call site for the km-distance conversion. Handlers now use it uniformly.
- **Why:** Single source of truth for a load-bearing constant. If the reference point or the 50km limit ever needs to change, there's one place to edit it.

### 2. Collapse the five repeated `tools/call` arms into a helper (`src/mcp/server.rs`)

- **Before:** Each of the five tool dispatch arms was an 8-line block repeating `serde_json::from_value -> handler.<tool>(input).await? -> json!({"content": [{"type": "text", "text": to_string_pretty(&output)?}]})`.
- **After:** A small generic `tool_text_content<I, O, _, _>(arguments, call)` helper encapsulates that pipeline. Each arm is now a single `tool_text_content(arguments, |input| handler.<tool>(input)).await` call.
- **Why:** ~40 lines of duplicated serialization/envelope code removed. Adding a new MCP tool now touches exactly one line in the dispatch table instead of pasting an 8-line block.

### 3. Collapse the four repeated resource error arms (`src/mcp/server.rs`)

- **Before:** Each of the four `handle_resource` branches duplicated the same `error!("Failed to get ...: {}", e); (StatusCode::INTERNAL_SERVER_ERROR, Json(json!({"error": ..., "details": e.to_string()}))).into_response()` pattern.
[...]

PR #110, opened 2026-04-21
arch: parser tests + panic fix, validation helper, dedupe resource dispatch

DominicBurkart:
## Summary

Architecture-alignment pass on the data + MCP layers. Three independent cleanups on this branch, all keeping behavior (one error variant corrected). 135 tests pass; `cargo clippy -D warnings` and `cargo fmt --check` clean.

### `src/data/client.rs` — parser unit tests + missing-lat/lon panic fix (commit `7fd4e50`)

`parse_reference_station` and `parse_realtime_status` had zero direct coverage — they were reachable only through the live-API-only `fetch_*` methods. That means upstream schema drift would silently drop records via `if let Ok(station) = ...` with no CI signal.

- Promote both parsers from `fn(&self, ...)` to `fn(...)` (they never used `self`) so the adjacent `#[cfg(test)]` module can call them without a client instance.
- +20 unit tests pinning every required-field branch, status-derivation matrix, and `duedate` fallback.
- Fix latent panic: `geo_point["lat"]` indexes into a `serde_json::Map`, which panics on missing keys rather than returning the documented `Err`. Swap to `.get("lat").and_then(Value::as_f64)`. Two of the new tests regression-pin this.

Net: lib tests 64 → 84, no public-API change, no new deps.

### `src/mcp/handlers.rs` — consolidate service-area validation + correct error variant (commit `f53deb4`)

- Extract `validate_service_area(&Coordinates) -> Result<()>`. The Paris-metro check + 50 km service-area check was inlined once in `find_nearby_stations` and twice in `plan_bike_journey` (origin and destination), all emitting the same `InvalidCoordinates` / `OutsideServiceArea` errors. Collapses to one helper + three call sites.
- `search_stations_by_name` returned `Error::Internal(anyhow!("Search query too short"))` for a 1-character query, which maps to MCP `-32603 internal_error`. A short query is invalid client input, not an internal failure — switch to `Error::Validation(..)` so it surfaces as `-32602 invalid_params` / `validation_error`, matching the sibling input-validation branches and the `CLAUDE.md` "Gestion erreurs appropriée" precept.

### `src/mcp/server.rs` — dedupe `handle_resource` match arms (commit `f53deb4`)

Four copies of the same `match { Ok => Json, Err => error!+500 }` scaffold, one per resource URI. Factored into a single match that picks the future + subject string, then shares one success/error tail. Identical behavior.

### `docs` / `README` / `CLAUDE.md` — stale-content trim (commit `121c4a0`)

- README: fix Claude Code install command (`claude mcp add` instead of `claude config add-server`), stop conflating `cargo install` with `cargo run --release`.
- CLAUDE.md: drop the fabricated multi-agents-process section with invented "+25% quality, -40% cycle time" metrics; keep context, objectives, current state, file map, dev commands, deployment.
- `docs/context/etat_actuel.md`: refresh from "Phase 2 not started, 2025-06-14" to phases 0–4 done, matching CLAUDE.md.
- `docs/development/project_prompt.md`: remove `[... rest of the original prompt content ...]` placeholder + duplicate intro; point at CLAUDE.md as the single source of truth.

Aggregate: `+3xx / -1xx` LOC dominated by the new parser tests; arch cleanup on its own is -40 LOC.

## Test plan

- [x] `cargo test --lib` — 84 passing (from 64).
- [x] `cargo test --all` — 135 passing, no regressions.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] Pre-push hook (`cargo test --all`) passed.
- [ ] CI green on this branch.

https://claude.ai/code/session_017vrZ4D421Ty4uKSgmm7Qsb

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,227 +1,82 @@
 # Projet CLAUDE : Serveur MCP Velib
 
 ## Contexte et Rôle
-Tu es un développeur Rust expert travaillant sur un projet open-source de serveur MCP.
+Développeur Rust travaillant sur un projet open-source de serveur MCP.
 
-- **Répertoire de travail** : `~/code/velib-mcp/velib-mcp` (main repo)
-- **Architecture worktree** : Branches adjacentes (`~/code/velib-mcp/branch1/`, etc.)
 - **Outils disponibles** : git, cargo, podman, CLI scw, CLI gh
-- **Public cible** : Assistants IA nécessitant l'accès aux données Velib
-- **Durée prévue** : Projet de développement multi-jour
-- **Contexte** : Développement collaboratif avec possibilité de travail parallèle sur différents worktrees
-
-### Structure du Projet
-```
-~/code/velib-mcp/
-├── velib-mcp/              # Dépôt principal (ce répertoire)
-│   ├── CLAUDE.md           # Configuration Claude partagée
-│   ├── src/                # Code source
-│   ├── docs/               # Documentation
-│   └── ...                 # Fichiers du projet
-├── branch1/                # Worktree pour branche feature
-│   ├── CLAUDE.md -> ../velib-mcp/CLAUDE.md  # Symlink vers config principale
-│   └── ...                 # Fichiers spécifiques à la branche
-└── branch2/                # Autre worktree
-    ├── CLAUDE.md -> ../velib-mcp/CLAUDE.md  # Symlink vers config principale
-    └── ...                 # Fichiers spécifiques à la branche
-```
-
-**Important** : Ce fichier CLAUDE.md est partagé via symlinks vers tous les worktrees pour maintenir un contexte cohérent.
+- **Public cible** : assistants IA nécessitant l'accès aux données Velib
+- **Collaboration** : travail parallèle possible via `git worktree`
 
 ## Objectif du Projet
-Créer un serveur cloud MCP performant pour rendre accessibles aux assistants IA les deux jeux de données parisiens suivants :
+Serveur cloud MCP rendant accessibles aux assistants IA les deux jeux de
+données Velib de la ville de Paris :
 
-- **Disponibilité temps réel** : https://opendata.paris.fr/explore/dataset/velib-disponibilite-en-temps-reel/information/?disjunctive.is_renting&disjunctive.is_installed&disjunctive.is_returning&disjunctive.name&disjunctive.nom_arrondissement_communes
+- **Disponibilité temps réel** : https://opendata.paris.fr/explore/dataset/velib-disponibilite-en-temps-reel/information/
 - **Emplacements des stations** : https://opendata.paris.fr/explore/dataset/velib-emplacement-des-stations/information/
 
-- **But** : Rendre toute information possible de ces jeux de données accessible aux assistants IA pour la planification des transports et l'analyse des flux de trajets.
+**But** : exposer l'intégralité des informations de ces datasets pour la
+planification de transport et l'analyse des flux.
 
 ## État Actuel du Projet
 
-### Phases Terminées ✅
-- **Phase 0** : Configuration projet, CI/CD, structure documentation
-- **Phase 1** : Analyse complète des données Velib (15+ champs documentés)
-- **Phase 2A** : Configuration environnement et fondation serveur de base
-- **Phase 2B** : Fondation protocole MCP et types de base
-- **Phase 3A** : Intégration API live et client de données
-- **Phase 3B** : Handlers MCP complets avec intégration données live
-- **Phase 4** : Nettoyage structure repository (suppression worktrees committés)
+### Phases terminées
+- **Phase 0** : configuration projet, CI/CD, structure documentation
+- **Phase 1** : analyse des données Velib (15+ champs documentés)
+- **Phase 2A** : environnement et fondation serveur
+- **Phase 2B** : protocole MCP et types de base
+- **Phase 3A** : intégration API live et client de données
+- **Phase 3B** : handlers MCP avec données temps réel
+- **Phase 4** : nettoyage structure repository
 
-### Architecture Technique
-- **Serveur MCP Rust** pour données Velib Paris
-- **Deux datasets principaux** :
-  - Disponibilité stations en temps réel
-  - Localisations et métadonnées des stations
-- **Déploiement Scaleway** via GitHub Actions
-- **Suite de tests complète** (18+ tests)
-- **Validations sécurité** incluant limites zone service 50km
+Voir `docs/context/etat_actuel.md` pour l'historique détaillé.
 
-### Fichiers Importants
-- `/src/main.rs` - Point d'entrée principal
-- `/src/mcp/` - Implémentation protocole MCP
-- `/src/data/` - Client données et cache
-- `/src/types.rs` - Structures de données principales
-- `/docs/api/data_analysis.md` - Analyse données complète
-- `/docs/context/etat_actuel.md` - Suivi statut projet
+### Architecture technique
+- Serveur MCP en Rust pour données Velib Paris
+- Deux datasets : disponibilité temps réel + métadonnées stations
+- Déploiement Scaleway Container Serverless via GitHub Actions
+- Suite de tests (18+ tests)
+- Validations sécurité dont limite zone de service 50 km
 
-### Commandes Développement
+### Fichiers importants
+- `src/main.rs` — point d'entrée
+- `src/mcp/` — implémentation MCP
+- `src/data/` — client données et cache
+- `src/types.rs` — structures principales
+- `docs/api/data_analysis.md` — analyse des données
+- `docs/api/mcp_interface_spec.md` — spécification MCP
+- `docs/api/mcp_schemas.md` — schémas Rust
+
+## Commandes Développement
 ```bash
-cargo test                     # Tests complets
-cargo fmt                      # Formatage code
-cargo clippy                   # Analyse statique
-cargo audit                    # Audit sécurité
+cargo test       # Tests
+cargo fmt        # Formatage
+cargo clippy     # Lints
+cargo audit      # Audit sécurité
 ```
 
-### Déploiement
+Les hooks pre-commit (`cargo-husky`) sont détaillés dans `README.md` section
+« Pre-commit hooks ».
+
+## Déploiement
 - **Cible** : Scaleway Container Serverless
-- **Déclencheur** : Push vers branche main
+- **Déclencheur** : push sur `main`
 - **Registry** : Scaleway Container Registry
-- **Build** : Containerisation Podman
+- **Build** : image distroless construite avec Podman
 
-### Gestion Worktrees
+## Worktrees
 ```bash
-# Créer nouveau worktree
+# Créer un worktree
 git worktree add ../branch-name branch-name
-cd ../branch-name
-ln -s ../velib-mcp/CLAUDE.md CLAUDE.md
 
-# Supprimer worktree
+# Supprimer un worktree
 git worktree remove ../branch-name
 git worktree prune
 ```
 
-## Processus de Développement Multi-Agents
+## Processus de Développement
 
-### Architecture Optimisée pour Performance, Qualité et Autonomie
+TDD : écrire les tests avant l'implémentation. À chaque commit, le hook
+pre-commit exécute `cargo clippy --fix`, `cargo fmt`, `cargo sort`, et
+`cargo deny check` — un échec annule le commit.
 
-Ce processus transforme l'approche linéaire traditionnelle en 5 phases parallèles pour maximiser l'efficacité d'équipe.
-
-#### Phase 1: Analyse Concurrente (PM + Test Designer)
-**Durée**: ~30 min | **Parallélisation**: PM et Test Designer travaillent simultanément
-
-**Product Manager (Rôle: Extraction de Valeur)**
-- Analyse issue GitHub avec template structuré
-- Extraction valeur métier claire et actionnable
-- Validation exigences avec dev-utilisateur
-- Production: Spécification fonctionnelle validée
-
-**Test Designer (Rôle: Planification Technique)**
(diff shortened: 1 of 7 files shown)


src/data/client.rs
```diff
@@ -326,3 +332,309 @@ impl VelibDataClient {
...
+    // --- parse_realtime_status: numeric fields cast silently ---
+
+    #[test]
+    fn realtime_numeric_fields_truncate_to_u16() {
+        // Values larger than u16::MAX are cast via `as u16` (wrap). The
+        // upstream API never produces such values, but pinning the behavior
+        // makes any future switch to checked conversion a deliberate decision.
+        let record = json!({
+            "stationcode": "1",
+            "mechanical": 70_000u64,
```
DominicBurkart, line 635: Test name claims truncation but only asserts `is_ok()`. Either assert the wrapped value (e.g. `70_000 as u16 == 4464`) so the contract is actually pinned, or rename to `realtime_numeric_fields_accept_oversized_values`. As written a future switch to checked conversion would still pass this test.

---

src/data/client.rs
```diff
@@ -326,3 +332,309 @@ impl VelibDataClient {
         (reference_size, realtime_size)
     }
 }
+
+#[cfg(test)]
```
DominicBurkart, line 336: Parser coverage is solid for known branches, but everything is example-based. Consider a `proptest` (or `cargo fuzz`) target over arbitrary `serde_json::Value` for both parsers: the whole point of these tests is catching upstream schema drift, and hand-written cases only cover drifts you already imagined. A single property ("parser never panics on arbitrary JSON, only returns Err") would directly regression-guard the class of bug you just fixed.

---

src/data/client.rs
```diff
@@ -238,17 +238,23 @@ impl VelibDataClient {
...
 
-        // Parse coordinates from coordonnees_geo
+        // Parse coordinates from coordonnees_geo.
+        // Use `.get(...)` rather than `geo_point["lat"]`: indexing into a
+        // `serde_json::Map` panics on missing keys, so we need an explicit
+        // lookup to surface a clean `Err` when the upstream payload is
+        // missing `lat`/`lon`.
         let geo_point = record["coordonnees_geo"]
             .as_object()
             .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing geo coordinates")))?;
```
DominicBurkart, line 249: While you're here: `record["stationcode"]`, `record["name"]`, `record["capacity"]`, and `record["coordonnees_geo"]` above all index into `record` the same way the old `geo_point["lat"]` did. For a top-level `serde_json::Value` that's `Value::Null` on miss rather than a panic, so it's not a live bug, but the inconsistency with the `.get(...).and_then(...)` style you just introduced is a footgun — if `record` ever becomes a `Map` in a refactor, the panics come back. Worth normalizing the whole parser to `.get(...)` for uniformity.

---

src/mcp/handlers.rs
```diff
@@ -23,6 +23,23 @@ const PARIS_CITY_HALL: Coordinates = Coordinates {
...
+    if !point.is_valid_paris_metro() {
+        return Err(Error::InvalidCoordinates {
+            latitude: point.latitude,
+            longitude: point.longitude,
+        });
+    }
+    if !point.is_within_paris_service_area() {
+        let distance_km = point.distance_to(&PARIS_CITY_HALL) / 1000.0;
+        return Err(Error::OutsideServiceArea { distance_km });
+    }
```
DominicBurkart, line 39: Helper is reused at all three call sites and preserves error variants exactly — dedupe goal met. One small tweak: there's still no direct unit test for `validate_service_area` itself. Given it's now the single chokepoint for service-area policy, a handful of tests pinning (valid Paris point, out-of-metro, out-of-service-radius) would protect the invariant independently of the handler call graph.

---

src/mcp/server.rs
```diff
@@ -445,74 +445,48 @@ async fn handle_resource(
...
+
+    match result {
+        Ok(response) => Json(response).into_response(),
+        Err(e) => {
+            error!("Failed to get {}: {}", subject, e);
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(json!({
+                    "error": format!("Failed to fetch {subject}"),
+                    "details": e.to_string()
```
DominicBurkart, line 485: Dispatch refactor is correct — all four URIs still route to distinct futures, the `_` → 404 branch is preserved, and the success/error tail is genuinely shared. No accidental route collapse. Nit: the new error payload uses `format!("Failed to fetch {subject}")`, which changes the wire message from e.g. `"Failed to fetch real-time stations"` (previous literal) to the same string built dynamically — identical output, just worth noting if anything downstream pattern-matches on exact bytes.

---

CLAUDE.md
```diff
@@ -1,227 +1,82 @@
...
+pre-commit exécute `cargo clippy --fix`, `cargo fmt`, `cargo sort`, et
+`cargo deny check` — un échec annule le commit.
 
-Ce processus s'intègre parfaitement avec:
-- Architecture worktree existante (isolation parallèle)
-- CI/CD GitHub Actions (validation automatisée)
-- Hooks pre-commit (qualité continue)
-- Toolchain Rust standard (fmt, clippy, audit)
\ No newline at end of file
+Chaque PR référence son issue d'origine et passe la CI complète avant merge.
```
DominicBurkart, line 82: Good call dropping the fabricated "+25% quality / -40% cycle time" section — those numbers were invented and would mislead future agents. The trimmed file is now a genuine repo-context doc rather than a process-theater artifact. No objection from a docs-bloat angle; this is a net reduction.

---

DominicBurkart (commented):
Three independent cleanups, all behavior-preserving where claimed.

**Panic fix**: genuinely removed, not moved — `.get().and_then()` returns `Err` and two regression tests pin it. Same-shape `record["x"]` accesses elsewhere are `Value::Null`-on-miss (not panics), so not a live bug, but worth normalizing for consistency (see inline).

**Validation helper**: reused at all 3 call sites, error variants and payloads match the originals exactly. Dedupe goal achieved. Missing: direct unit tests for the helper itself.

**Parser tests**: +20 solid example-based tests across required-field, status-derivation matrix, and duedate fallback. Gap: no property-based / fuzz coverage, which is exactly what would catch the class of bug you just fixed (upstream schema drift). Worth a follow-up proptest.

**Dispatch dedupe**: correct — 4 URIs still route distinctly, 404 branch preserved, shared tail is behaviorally identical.

**Docs**: trimming invented metrics and stale phase status is a clean win. No secrets, no bloat.

No blockers. See inline for nits and follow-up suggestions.

---

DominicBurkart:
## CI triage — 2026-04-21

**Checks:** all 8 green (Rustfmt, Security Audit, Test Suite, Feature Testing, Cargo Sort, Clippy, Build, All Checks Passed).
**Blocker:** `mergeable_state: dirty` — real textual conflict with `main` in `src/mcp/handlers.rs` and `src/mcp/server.rs`.

**Root cause:** PR #107 (`d08b47e refactor(mcp): deduplicate tools/call and resource handler branches`, `2cd9553 refactor(handlers): use Coordinates helper for service-area distance`) landed on main and already delivered the arch cleanups this PR proposes — an `ensure_in_service_area` helper (same body, different name from my `validate_service_area`) and the `handle_resource` dedupe.

**Sibling PRs:** #111, #113 `behind` main; #112, #114 `blocked`. Not a shared-CI regression — each has its own base-branch freshness issue.

**Rebase outcome (local, unpushed):**
- `docs:` commit → clean apply.
- `test:` parser tests + panic fix → clean apply, retains all value.
- `arch:` commit → shrinks to the single net-new delta: `Error::Internal` → `Error::Validation` in `search_stations_by_name`. Helper + server dedupe dropped (already on main).
- `cargo build --lib`: clean. `cargo test --lib`: 87 passed, 0 failed.

**Why not pushed:** this sandbox has no git push credentials (HTTPS prompts for user; no SSH binary; no `gh`) and local commit signing also fails (`signing server returned status 400: missing source`). GitHub API `update-branch` returns 422 (real conflict, not fast-forward).

**Suggested next step for a maintainer:**
```
git fetch origin
git rebase origin/main
# take --ours for both conflicts (main already has equivalent refactors),
# then re-apply the Error::Validation fix:
#   src/mcp/handlers.rs search_stations_by_name: Internal → Validation("Search query must be at least 2 characters")
git push --force-with-lease
```

Or, given the arch commit is now largely redundant, consider squashing to just the docs + parser-tests + one-line validation-variant fix.

Status: **BLOCKED on environment (no push auth / no signing)** — not fixable from this agent session.

---

DominicBurkart:

Review posted with 6 inline comments: panic truly removed via `.get().and_then()` + 2 regression tests; same-shape `record["x"]` accesses elsewhere are Null-on-miss (inconsistent, not panicking); `validate_service_area` reused at all 3 call sites but no direct unit test; dispatch refactor preserves all 4 routes + 404; parser tests example-based only (no fuzz/proptest); one test (`realtime_numeric_fields_truncate_to_u16`) doesn't actually assert truncation. `mergeable_state: dirty` — main's PR #107 already shipped equivalent validation-helper + resource-dispatch work, superseding most of this arch commit. CI-fix agent cloned, rebased on origin/main, resolved conflicts (taking main's versions + re-applying the unique `Error::Validation` delta in `search_stations_by_name`), and verified `cargo build --lib` + `cargo test --lib` 87/87 passing locally — but **push BLOCKED** by environment (no credentials, signing-server 400, API `update-branch` 422); diagnosis + rebase plan posted at [issuecomment-4292554303](https://github.com/DominicBurkart/velib-mcp/pull/110#issuecomment-4292554303).

---

DominicBurkart (commented):
## Code review — PR #110

Three independent cleanups, all behavior-preserving as described. No blockers. Notes below are optional follow-ups; would approve if I weren't the author.

### Panic fix (`src/data/client.rs`) — correct root-cause fix

`serde_json::Map` indexing panics on missing keys (`Index for Map` — `index not found`), so the old `geo_point["lat"].as_f64()` would panic whenever upstream dropped a coordinate field and `fetch_*`'s `if let Ok` would never even be consulted. The switch to `geo_point.get("lat").and_then(Value::as_f64)` addresses the root cause (not a symptom): the `Option` flows naturally into `ok_or_else(...)` and the `Err` branch the code already claimed to produce. Two regression tests (`reference_missing_latitude_errors`, `reference_missing_longitude_errors`) pin it. Good fix.

Minor consistency note (non-blocking): the surrounding `record["stationcode"]`, `record["coordonnees_geo"]` accesses *look* similar but return `Value::Null` (they index `Value`, not `Map`), so they are not panic hazards — they rely on `.as_str()` / `.as_object()` returning `None`. Worth a follow-up normalization pass to `.get(...)` everywhere for readability, but not a correctness issue.

### Parser tests — strong coverage

The +20 tests meaningfully close the gap (happy path, every required-field branch, wrong-type rejection, status-derivation matrix including `"unknown"` as closed, duedate RFC3339 + unparseable + missing fallback, u16 truncation). The module docstring explicitly documents the invariants being pinned, which is the right framing for schema-drift canaries. One follow-up worth considering: property-based tests (proptest/arbitrary over `Value`) would be the natural next layer — exactly the class of bug the panic fix addressed.

### Validation helper (`validate_service_area`)

Correctness preserved: at all three call sites the extracted helper emits the same `Error::InvalidCoordinates { latitude, longitude }` and `Error::OutsideServiceArea { distance_km }` values, with the same branch ordering and the same `distance_to(&PARIS_CITY_HALL) / 1000.0` math. Scope is right — the helper captures a real duplication (the Paris-metro + 50 km check was pasted 3×) and doesn't over-generalize. No direct unit tests for the helper (it's covered transitively through handler tests); a couple of focused tests on the helper function itself would make future changes safer.

### `Error::Internal` → `Error::Validation` for short query

Correct reclassification per the PR description. A 1-char query is invalid client input, and this aligns with the sibling `MAX_RESULT_LIMIT` branch that already uses `Error::Validation`. Worth adding a regression test asserting the JSON-RPC error code is `-32602` / `validation_error` for a 1-char query so this can't silently regress to `-32603`.

### `handle_resource` dedupe (`src/mcp/server.rs`)

Correctness preserved: all 4 URIs still dispatch to distinct futures, the 404 branch is preserved via early `return`, and the shared success/error tail produces the same status code (500) and the same JSON shape (`{error, details}`) as before. The subject-string pattern (`"reference stations"` → `"Failed to fetch reference stations"`) reproduces the previous messages verbatim. Dedupe goal achieved, no behavior drop.

### Overlap with PR #114 (`describe_api`) — merge-order concern

**Heads-up**: #114 is open against `main` and independently refactors the same `handle_resource` function — it introduces its own `resource_error(err, user_message)` helper (a *different* refactor shape than #110's `(result, subject)` tuple match) **and** adds a 5th resource URI `velib://api/description`. A textual merge conflict in `src/mcp/server.rs::handle_resource` is near-certain regardless of merge order. No semantic conflict (the two refactors are compatible intent-wise) but the second-to-merge PR will need a manual rebase. Worth coordinating — the #110 tuple-match style composes more cleanly with a 5th resource than the #114 per-branch `resource_error` style, so merging #110 first and rebasing #114 on top is the lower-cost order. (Separately, #114 also promotes `REFERENCE_CACHE_TTL_MINUTES` / `REALTIME_CACHE_TTL_MINUTES` to `pub const` — no conflict with #110 but worth noting.)

### Docs trim

No opinion on the French editorial changes; removing the fabricated "-40% cycle time / +25% quality" metrics and the stale "Phase 2 not started, 2025-06-14" state is a clear win. README install command correction is accurate (`claude mcp add` is the current CLI form). No secrets or generated artifacts committed.

---

**Recommendation**: ready to merge. Suggested follow-ups (none blocking): normalize the remaining `record["..."]` accesses to `.get(...)`, add direct tests for `validate_service_area`, add a test pinning `-32602` for the short-query path, coordinate merge order with #114 (merge #110 first). `mergeable_state` is currently `dirty` against main — rebase needed before merge.

---

PR #111, opened 2026-04-21
fix(deny): allow Unicode-3.0, bump rand, ignore rustls-pemfile advisory

DominicBurkart:
## Summary

`cargo deny check` is currently failing against `main` for three reasons; CI does not run deny, so this went unnoticed. Fixes below.

- **Licenses**: newer `icu_*` crates (pulled via `reqwest` → `url` → `idna`) ship under `Unicode-3.0`. Added to the license allow list alongside `Unicode-DFS-2016`.
- **RUSTSEC-2026-0097** (`rand` 0.8.5 unsoundness): fixed by bumping `rand` to `0.8.6` via `cargo update -p rand --precise 0.8.6`. Semver-compatible, no manifest change.
- **RUSTSEC-2025-0134** (`rustls-pemfile` unmaintained): transitive via `reqwest` 0.11. The only real fix is bumping to `reqwest` 0.12, which is a larger lift. Ignored with a documented reason for now.

Unblocks the pre-commit hook, which currently fails on every commit that goes through it.

## Test plan

- [x] `cargo deny check` passes locally (`advisories ok, bans ok, licenses ok, sources ok`)
- [ ] CI green on this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1027,9 +1027,9 @@ checksum = "74765f6d916ee2faa39bc8e68e4f3ed8949b48cccdac59983d287a7cb71ce9c5"
 
 [[package]]
 name = "rand"
-version = "0.8.5"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
+checksum = "5ca0ecfa931c29007047d1bc58e623ab12e5590e8c7cc53200d5202b69266d8a"
 dependencies = [
  "libc",
  "rand_chacha",
--- a/deny.toml
+++ b/deny.toml
@@ -1,3 +1,8 @@
+[advisories]
+ignore = [
+  { id = "RUSTSEC-2025-0134", reason = "rustls-pemfile unmaintained; transitive via reqwest 0.11. No safe upgrade without bumping reqwest to 0.12 (tracked separately)." },
+]
+
 [licenses]
 allow = [
   "MIT",
@@ -7,6 +12,7 @@ allow = [
   "BSD-2-Clause",
   "BSD-3-Clause",
   "Unicode-DFS-2016",
+  "Unicode-3.0",
   "CC0-1.0",
   "Zlib",
   "OpenSSL",


DominicBurkart (commented):
Review for coding agents.

**Overlap with sibling PRs**: This PR is effectively a **superset** of #113 (Unicode-3.0 license only) and an **alternative** to #112 (which bumps `reqwest` 0.11 -> 0.12, transitively dropping `rustls-pemfile` and fixing RUSTSEC-2025-0134 at its root). Only one of {#111, #112+#113} should land:

- If #112 merges first, the `RUSTSEC-2025-0134` ignore in this PR becomes dead config (rustls-pemfile is gone) and the `rand` 0.8.6 bump here is still useful but #112 already picks up 0.8.6 via `reqwest` 0.12's tree.
- If this PR merges first, #112 still needs to land to actually remove the unmaintained dependency; the ignore here is explicitly a stopgap.
Recommend closing #113 as redundant and picking either (a) this as a fast unblock + follow-up with #112, or (b) #112 directly and drop the ignore entry.

**Findings:**

1. `RUSTSEC-2025-0134` ignore lacks a `expiration` / end-date and doesn't link a tracking issue. `cargo-deny` supports `expiration = "YYYY-MM-DD"` on ignore entries; without it this silently becomes permanent. Body says "tracked separately" but no issue number is referenced.
2. `rand` 0.8.5 -> 0.8.6 is `Cargo.lock`-only and semver-compatible; no API surface changed that affects this crate (uses `fastrand` as a direct dep; `rand` is transitive via `tokio-tungstenite`). Low breakage risk, acceptable.
3. `Unicode-3.0` addition is correctly scoped (license-list-only, alongside existing `Unicode-DFS-2016`); same change as #113.
4. PR is `draft` + `mergeable_state: behind` - needs rebase on `main` before merge.
5. Per `CLAUDE.md`, repo expects `cargo audit` in dev loop but CI doesn't run `cargo deny` (PR body confirms this). Consider a follow-up to add `cargo deny check` to CI so this class of drift is caught.

---

deny.toml
```diff
@@ -1,3 +1,8 @@
+[advisories]
+ignore = [
+  { id = "RUSTSEC-2025-0134", reason = "rustls-pemfile unmaintained; transitive via reqwest 0.11. No safe upgrade without bumping reqwest to 0.12 (tracked separately)." },
```
DominicBurkart, line 3: Missing `expiration` field. Without an end-date this ignore is effectively permanent and will hide the advisory long after #112 lands and removes `rustls-pemfile`. Suggest:

```toml
{ id = "RUSTSEC-2025-0134", reason = "rustls-pemfile unmaintained; transitive via reqwest 0.11. Tracked by #<issue>; resolved when reqwest bumps to 0.12 (see PR #112).", expiration = "2026-07-21" }
```

Also: the `reason` claims the upgrade is "tracked separately" but doesn't reference PR #112 or an issue number - please add one so agents triaging this can follow the thread.

---

deny.toml
```diff
@@ -7,6 +12,7 @@ allow = [
   "BSD-2-Clause",
   "BSD-3-Clause",
   "Unicode-DFS-2016",
+  "Unicode-3.0",
```
DominicBurkart, line 15: `Unicode-3.0` addition looks correct and matches #113 exactly. Scope is minimal (license allowlist only, no clarifications added) - ICU crates ship under this SPDX identifier since Unicode CLDR 44, so this is accurate. Consider closing #113 as redundant once this merges.

---

Cargo.lock
```diff
@@ -1027,9 +1027,9 @@ checksum = "74765f6d916ee2faa39bc8e68e4f3ed8949b48cccdac59983d287a7cb71ce9c5"
 
 [[package]]
 name = "rand"
-version = "0.8.5"
+version = "0.8.6"
```
DominicBurkart, line 1030: `rand` 0.8.5 -> 0.8.6 is a patch bump and `rand` is only a transitive dep here (via `tokio-tungstenite`); the crate itself uses `fastrand` directly per `Cargo.toml`. Semver-compatible, no public API changes between these versions. No breakage risk. RUSTSEC-2026-0097 is correctly resolved at the root rather than ignored.

---

DominicBurkart (commented):
Inline notes on the three hunks. See top-level review for overlap analysis with #112/#113.

---

DominicBurkart:

Closed as superseded by `main`: commit `610bc6b` already ships the identical diff (rand 0.8.6 bump + Unicode-3.0 allow + RUSTSEC-2025-0134 ignore). CI on `89b660a` was fully green; `mergeable_state=behind` solely because another merge path landed the same content — rebasing would yield an empty commit. Review (3 inline comments on advisory expiration ref, Unicode-3.0 scope, rand transitive-only) is retained on the PR for future reference.

---

PR #112, opened 2026-04-21
chore(deps): bump reqwest to 0.12 and rand to 0.8.6

DominicBurkart:
## Summary

- **`reqwest`** `0.11` → `0.12`: drops the transitive dep on the now-archived `rustls-pemfile` crate (RUSTSEC-2025-0134). No source-level changes needed — velib-mcp only uses `reqwest::{Client, Response, Error}`, all API-compatible.
- **`rand`** `0.8.5` → `0.8.6` (lockfile-only): addresses RUSTSEC-2026-0097 (unsound with custom logger). Pulled in transitively via `axum → tokio-tungstenite → tungstenite`.

## Coupling note

`reqwest 0.12` brings ICU 2.0 crates into the tree, which are licensed `Unicode-3.0`. The sibling PR **`chore(deny): allow Unicode-3.0 license`** adds that to `deny.toml`'s allowlist; the two should land together or the license one first.

Local `.git/hooks/pre-commit` was bypassed once (with user authorization) because the hook runs both checks together — creating a chicken-and-egg with the license PR. CI will run cargo-deny cleanly once both are on main.

## Test plan

- [x] `cargo build` clean
- [x] `cargo test --lib` — 67 passed
- [x] `cargo deny check advisories` — ok
- [ ] CI green on this PR
- [ ] CI green after license PR merges

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,6 +59,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -72,13 +78,13 @@ checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
  "axum-core",
- "base64 0.22.1",
+ "base64",
  "bytes",
  "futures-util",
- "http 1.3.1",
- "http-body 1.0.1",
+ "http",
+ "http-body",
  "http-body-util",
- "hyper 1.6.0",
+ "hyper",
  "hyper-util",
  "itoa",
  "matchit",
@@ -92,7 +98,7 @@ dependencies = [
  "serde_path_to_error",
  "serde_urlencoded",
  "sha1",
- "sync_wrapper 1.0.2",
+ "sync_wrapper",
  "tokio",
  "tokio-tungstenite",
  "tower",
@@ -110,13 +116,13 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures-util",
- "http 1.3.1",
- "http-body 1.0.1",
+ "http",
+ "http-body",
  "http-body-util",
  "mime",
  "pin-project-lite",
  "rustversion",
- "sync_wrapper 1.0.2",
+ "sync_wrapper",
  "tower-layer",
  "tower-service",
  "tracing",
@@ -134,26 +140,14 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
- "windows-targets 0.52.6",
+ "windows-targets",
 ]
-
-[[package]]
-name = "base64"
-version = "0.21.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
 name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
-
-[[package]]
-name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
@@ -429,16 +423,16 @@ checksum = "07e28edb80900c19c28f1072f2e8aeca7fa06b23cd4169cefe1af5aa3260783f"
 
 [[package]]
 name = "h2"
-version = "0.3.26"
+version = "0.4.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "81fe527a889e1532da5c525686d96d4c2e74cdd345badf8dfef9f6b39dd5f5e8"
+checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
 dependencies = [
+ "atomic-waker",
  "bytes",
  "fnv",
  "futures-core",
  "futures-sink",
- "futures-util",
- "http 0.2.12",
+ "http",
  "indexmap",
  "slab",
  "tokio",
@@ -454,17 +448,6 @@ checksum = "5971ac85611da7067dbfcabef3c70ebb5606018acd9e2a3903a0da507521e0d5"
 
 [[package]]
 name = "http"
-version = "0.2.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "601cbb57e577e2f5ef5be8e7b83f0f63994f25aa94d673e54a92d5c516d101f1"
-dependencies = [
- "bytes",
- "fnv",
- "itoa",
-]
-
-[[package]]
-name = "http"
 version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f4a85d31aea989eead29a3aaf9e1115a180df8282431156e533de47660892565"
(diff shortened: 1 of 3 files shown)


DominicBurkart (commented):
Review for coding agents.

**Scope / overlap with siblings**
- Overlaps with #111 and #113 on the `rand` 0.8.5 -> 0.8.6 lockfile bump (all three PRs carry the identical change). Whichever lands first will make the others non-conflicting on that chunk, but expect a lock rebase. Consider dropping the rand bump from this PR and letting #111 or #113 own it, to keep #112 strictly a reqwest upgrade.
- #111 is mutually exclusive with #112: #111 ignores RUSTSEC-2025-0134 and allows `Unicode-3.0`, while #112 fixes the advisory at its source. Pick one strategy; if #112 lands, #111 becomes obsolete (the ignore entry should then be removed from `deny.toml`).
- #113 is a clean prerequisite for #112. Recommend landing #113 first so CI on #112 stays green without a hook bypass.

**API breakage (reqwest 0.11 -> 0.12)**
Source touchpoints are only `reqwest::{Client, Response, Error}` in `src/data/retry.rs` and `src/error.rs` per the PR description and search. The main 0.12 breaking changes that could bite:
- `Response::remote_addr()` -> now returns `Option<SocketAddr>` (was always present); not used here.
- Default TLS backend switched (still selectable via features). The Cargo.toml keeps only `features = ["json"]`, so reqwest 0.12 uses its default-tls (native-tls) -- Cargo.lock confirms `native-tls`+`hyper-tls 0.6` still present. Good: no behavioral TLS swap.
- `Error` variants reshuffled; if any code matches on `reqwest::Error` kind() this would break. Worth a grep-confirm in retry.rs before merging.
- `Response::cookies()` signature tightened (not used here).

**Bloat / tree impact**
- Cargo.lock now pulls in `rustls`, `rustls-webpki`, `rustls-pki-types`, `ring`, `tokio-rustls`, `hyper-rustls`, `tower-http`, `iri-string`, `untrusted`, `zeroize`, `subtle`, `atomic-waker`, `windows-registry`. That's a meaningful compile-time and binary-size cost for a crate that (per the PR body) doesn't actually need rustls -- reqwest 0.12 enables `hyper-rustls` transitively via defaults even when only `json` is requested. Consider `default-features = false, features = ["json", "default-tls"]` to drop the rustls stack, or explicitly opt into rustls-only and drop native-tls. Either halves the new deps. Not blocking, but worth a follow-up issue.
- `tower-http` + `iri-string` arrive because reqwest 0.12 uses them for redirect URI handling; unavoidable with default features.

**rand compat**
0.8.5 -> 0.8.6 is a patch bump; the advisory (RUSTSEC-2026-0097) is logger-specific and does not affect this crate's usage. No source changes expected; none present. Confirmed.

**Missed test updates**
- `tests/smoke_test.rs` and `tests/integration_test.rs` both use reqwest per code search, but are not in the diff. The PR summary says no source-level changes needed; this appears to hold for Client/Response/Error but please confirm locally that `cargo test` (not just `--lib`) passes -- the test plan only lists `cargo test --lib` (67 passing). Run the integration + smoke suites before un-drafting.
- No test added covering the TLS/redirect path change, but given reqwest is only used via `Client::get`/`.json()` this is fine.

**Gaps vs repo context / docs**
- The PR body acknowledges a pre-commit hook bypass. There's no change to `deny.toml` here; the Unicode-3.0 allow is in #113. Document in the PR body the required merge order (#113 -> #112), and ensure the final state removes the RUSTSEC-2025-0134 ignore from #111 if that PR lands first.
- `mergeable_state: behind`. Rebase on main (and on #113 once merged) before leaving draft.
- No CHANGELOG/README update; the repo does not appear to maintain one, so this is fine.

**Recommendation**
Approve in principle once: (1) #113 is merged first or included; (2) `cargo test` (full, not `--lib` only) is run; (3) consider the default-features cleanup to avoid pulling the rustls tree unused; (4) confirm #111's advisory-ignore is not also merged.

---

DominicBurkart:

Review posted flagging: `rand` bump duplicated across #111/#112/#113; #112 and #111 are mutually-exclusive strategies (with #111 now landed on main, #112 is the correct path forward); reqwest 0.12 pulls the full rustls+tower-http tree unused (consider `default-features=false`); integration/smoke tests (which use reqwest) were only verified via `--lib`. Triggered `update_pull_request_branch` (new base `f411180`, new head `459ef14c`); at report time 6/8 checks green, Build + Coverage still running. Not yet ready — needs one more CI poll before promoting out of draft.

---

DominicBurkart:
Addressed the review in b59ca7c:

- **Bloat / rustls tree**: set `reqwest = { version = "0.12", default-features = false, features = ["json", "default-tls"] }` in both `[dependencies]` and `[dev-dependencies]` to drop the rustls + tower-http + ring transitive stack. native-tls path retained, matching prior 0.11 behavior.
- **deny.toml**: removed the RUSTSEC-2025-0134 ignore entry now that the advisory is fixed at source by the reqwest bump. Left a comment explaining why the ignore list is empty.
- **`reqwest::Error` kind-matching**: confirmed via read of `src/data/retry.rs` and `src/error.rs` — only `.status()` and `#[from] reqwest::Error` are used; no `.kind()` match patterns that would break on 0.12.
- **rand bump overlap**: left in place here since #111/#113 have not landed on `main` (deny.toml still has the RUSTSEC-2025-0134 ignore on main, and Cargo.toml on main is still at reqwest 0.11). This PR can land standalone.
- **Full `cargo test`** (not just `--lib`) and CI re-run: deferred to CI on this new commit.

---

DominicBurkart (commented):
## Code Review — reqwest 0.11 → 0.12 / rand 0.8.6

**Scope verified:** `Cargo.toml` (dep + dev-dep bump), `Cargo.lock` regeneration. Source unchanged.

### Findings

**API compatibility (OK).** Source usage is confined to `reqwest::{Client, Response, Error}` plus `Client::new`, `get`, `send`, `query`, `Response::headers/status/error_for_status` (see `src/data/retry.rs`, `src/error.rs`). All of these are stable across 0.11 → 0.12, so the "no source-level changes" claim in the PR body checks out. `reqwest::Error` `#[from]` in `Error::Http` still compiles since the type path is unchanged.

**Lockfile review (OK).**
- `reqwest 0.12.28`, `hyper 1.6.0`, `h2 0.4.13`, `hyper-tls 0.6.0`, `tokio-native-tls` retained — native-TLS transport preserved (default features on reqwest 0.12). Good: no silent swap to rustls-only that would break prod TLS.
- `rustls-pemfile` is **gone** from the tree — RUSTSEC-2025-0134 resolved as claimed.
- `rand 0.8.6` not visible in this diff (no `rand` stanza edits), so the lockfile-only bump presumably already landed via #111/rebase. PR body says so; verify final `Cargo.lock` on main post-merge contains `rand 0.8.6`.
- All new sources are `registry+https://github.com/rust-lang/crates.io-index` — no git/unknown sources introduced. Clean.
- New crates pulled in: `rustls 0.23`, `rustls-webpki`, `rustls-pki-types`, `ring 0.17.14`, `hyper-rustls 0.27`, `tokio-rustls 0.26`, `tower-http 0.6.8`, `iri-string`, `zeroize`, `subtle`, `untrusted`, `atomic-waker`, `windows-registry`. Well-known, reputable crates.

### Concerns

1. **Dead weight from default features.** reqwest 0.12 with default features drags in the full rustls + webpki + ring stack *and* native-tls *and* tower-http — even though the code only needs `json`. Consider `default-features = false, features = ["json", "default-tls"]` (or `"rustls-tls"`) to drop ~12 transitive crates and reduce attack surface / compile time. Non-blocking.
2. **Coupling with license PR.** PR body correctly flags that Unicode-3.0 must be in `deny.toml` before merge; `deny.toml` at HEAD already includes `Unicode-3.0`, so the ordering risk is real but the tree is self-consistent on this branch. Ensure the sibling license PR lands first (or squash-merge order is enforced) to avoid a broken main.
3. **Integration/smoke tests.** PR body states `cargo test --lib` was run (67 tests). `tests/smoke_test.rs` and `tests/integration_test.rs` both consume `reqwest`; CI's full "Test Suite" job is green per check_runs, so this is covered — just noting the PR description understates the test coverage.
4. **Codecov project check is "failure" (-3.98%).** Unrelated to this diff (patch coverage is 100%), appears to be a measurement artifact. Does not block correctness.

### Security

No secrets or PII in the diff. No suspicious registries/git sources. Advisory RUSTSEC-2025-0134 is resolved and the corresponding ignore in `deny.toml` can be removed in a follow-up (it still lists the advisory as ignored, but the dep is gone, so cargo-deny will no-op).

### Recommendation

**APPROVE (non-blocking comments).** Dep upgrade is surgical and API-compatible; advisory is fixed; CI is green; lockfile is clean. Merge *after* the Unicode-3.0 license PR. Consider the `default-features = false` optimization as a follow-up.

---

PR #113, opened 2026-04-21
chore(deny): allow Unicode-3.0 license

DominicBurkart:
## Summary

Add \`Unicode-3.0\` to \`deny.toml\`'s license allowlist. ICU crates (\`icu_collections\`, \`icu_locale_core\`, \`icu_normalizer\`, etc.) migrated from \`Unicode-DFS-2016\` to \`Unicode-3.0\` upstream, so cargo-deny currently rejects any dep graph that pulls them in.

## Coupling note

Sibling PR #112 (\`chore(deps): bump reqwest+rand\`) is what introduces ICU 2.0 into the tree. This PR is the license side of the same chicken-and-egg — local \`.git/hooks/pre-commit\` runs licenses + advisories together, so neither PR can pass the hook on its own. Pre-commit bypassed once (with user authorization). CI runs cargo-deny cleanly once both are on main.

## Test plan

- [x] \`cargo deny check licenses\` — ok
- [ ] CI green on this PR
- [ ] CI green after #112 merges

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -7,6 +7,7 @@ allow = [
   "BSD-2-Clause",
   "BSD-3-Clause",
   "Unicode-DFS-2016",
+  "Unicode-3.0",
   "CC0-1.0",
   "Zlib",
   "OpenSSL",


deny.toml
```diff
@@ -7,6 +7,7 @@ allow = [
   "BSD-2-Clause",
   "BSD-3-Clause",
   "Unicode-DFS-2016",
+  "Unicode-3.0",
```
DominicBurkart, line 10: This exact line (`"Unicode-3.0",`) is also added by PR #111, which additionally bumps `rand` to 0.8.6 and ignores RUSTSEC-2025-0134. PR #113 is a strict subset of #111. Recommend closing #113 (and likely #112 as well, since #111 covers the license + advisory fixes without requiring the larger reqwest 0.12 jump — unless the intent is specifically to drop `rustls-pemfile` rather than ignore it). Whichever lands first will cause the other to become a no-op/conflict.

---

DominicBurkart (commented):
Narrow, correct one-line change. Main concern: this PR is redundant with sibling PRs. #111 already contains this exact `Unicode-3.0` allow line (plus the rand bump and rustls-pemfile advisory ignore). #112 bumps reqwest to 0.12, which is what drags in the Unicode-3.0 ICU crates in the first place. Three overlapping PRs for one problem — recommend consolidating: either land #111 alone (covers license + advisory without the reqwest 0.12 lift), or land #113 + #112 and close #111. No bugs, no secrets, no bloat in the diff itself.

---

DominicBurkart:

Closed as redundant: the single `deny.toml` Unicode-3.0 allowlist change here is a strict subset of #111, and the same line has already landed on `main` via commit `610bc6b`. CI on this PR was green (8/8) but the change is now a no-op. No tracking issue needs to remain open — the work has shipped.

---

Issue #9, opened 2025-06-18
MCP documentation endpoint


DominicBurkart:
LLMs require additional context and prompting to be able to interact effectively with MCPs. Without context, they make naive and incorrect requests and develop confidently wrong beliefs. 

As an expert system architect with experience with contextual prompting, ultrathink about how to communicate critical features of this API to the models that leverage it. 

Pay particular attention to the idea of a self-documentation endpoint. An additional endpoint with a hard-coded response which provides a comprehensive schema of the endpoints, including key features such as data freshness (including caching), data formats, units for numerical values, and all values (with definitions) for enums. Request and response parameters should be typed and documented with this approach. 

This information should be provided in a standard format. With a default of JSON, consider if other formats would better serve model's needs.

Issue #10, opened 2025-06-18
Pagination loops against the public API should handle rate limits (e.g., HTTP 429) and include retry/backoff logic to improve robustness under heavy load or API throttling.


DominicBurkart:
Pagination loops against the public API should handle rate limits (e.g., HTTP 429) and include retry/backoff logic to improve robustness under heavy load or API throttling.

- https://github.com/DominicBurkart/velib-mcp/pull/4#issuecomment-2977963059
-  https://github.com/DominicBurkart/velib-mcp/pull/8#discussion_r2153732201_

PR #114, opened 2026-04-21
feat(mcp): add describe_api tool + velib://api/description resource (#9)

DominicBurkart:
Closes #9.

## Summary

Adds a self-documentation endpoint so MCP clients (LLMs) can discover the server schema at runtime: server info, service area, units for numerical values, per-data-class cache TTLs, all enum values with definitions, per-tool input/output schemas, and JSON-RPC error codes.

Exposed both as:
- **Tool** `describe_api` — optional `format: "json" | "markdown"` (default `json`).
- **Resource** `velib://api/description` — idiomatic MCP discovery path.

Payload imports TTL constants directly from `data::client` (promoted to `pub const`) so the docs cannot silently drift from the cache behavior.

## Drift guards (tests in `tests/mcp_describe_tests.rs`)

- Schema-parity: every tool exposed by `tools/list` has a matching entry in `describe_api.tools[*]` with identical `input_schema` (shared source via `tool_definitions()`).
- Enum-variant drift: iterates documented enum values and round-trips each through the corresponding Rust enum (`StationStatus`, `DataFreshness`, etc.).
- Markdown content: non-empty, starts with `# `, and mentions every tool name.
- Unknown `format` argument is rejected.
- Top-level shape: required keys present.

## Out of scope

- No new dependencies (no `insta`, no `criterion`).
- No behavioral changes to existing endpoints.

## Test plan

- [x] `cargo test` — 123 passed, 0 failed, 8 ignored (pre-existing network-gated).
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [ ] CI green on this PR (will confirm once checks run).

## Agent-job metadata (for janitor dedup)

- job-id: `cron-2026-04-21-one_track-2`
- oldest-issue-id: `one_track#2`
- this-issue: `velib-mcp#9`
- last-commit-sha: `409def906a84844507395dca131af2fe951b2826`

Per cron policy, this PR is **draft** and is NOT marked `ready-for-review` — planner jobs never promote. The janitor decides promotion.

### DIFF
--- a/src/mcp/describe.rs
+++ b/src/mcp/describe.rs
@@ -1,0 +1,500 @@
+//! Self-documentation for the Velib MCP API.
+//!
+//! This module is the **single source of truth** for the machine-readable
+//! description of the server's public contract that LLM clients discover at
+//! runtime: server info, service area, units, cache TTLs, enum definitions,
+//! per-tool input/output schemas, and error codes.
+//!
+//! The payload is assembled from:
+//! - Cache TTL constants imported from [`crate::data::client`] so the document
+//!   cannot drift from the actual cache configuration.
+//! - Per-tool `inputSchema` values taken from
+//!   [`crate::mcp::server::tool_definitions`] so the schemas reported here are
+//!   byte-for-byte the same ones returned by the `tools/list` JSON-RPC method.
+//! - Static metadata (enum variants, units, service-area bounds, error codes)
+//!   hard-coded here; any change to the corresponding Rust types must be
+//!   reflected here and is exercised by the drift-guard tests in
+//!   `tests/mcp_describe_tests.rs`.
+//!
+//! Both the `describe_api` tool (`tools/call`) and the
+//! `velib://api/description` resource (`resources/read`) render from the exact
+//! same JSON produced by [`api_description`].
+
+use serde_json::{json, Value};
+
+use crate::data::client::{REALTIME_CACHE_TTL_MINUTES, REFERENCE_CACHE_TTL_MINUTES};
+use crate::mcp::server::tool_definitions;
+use crate::types::{PARIS_CITY_HALL, PARIS_SERVICE_AREA_MAX_METERS};
+
+const SERVER_NAME: &str = "velib-mcp";
+const SERVER_VERSION: &str = env!("CARGO_PKG_VERSION");
+const MCP_PROTOCOL: &str = "2024-11-05";
+
+/// Build the full machine-readable API description.
+///
+/// Pure function: no I/O, no allocations that depend on runtime state beyond
+/// formatting the returned JSON value. Safe to call from any context.
+#[must_use]
+pub fn api_description() -> Value {
+    json!({
+        "server": {
+            "name": SERVER_NAME,
+            "version": SERVER_VERSION,
+            "protocol": MCP_PROTOCOL,
+            "endpoints": {
+                "http_jsonrpc": "POST /mcp",
+                "websocket_jsonrpc": "GET /mcp/ws (upgrade)",
+                "resource_http": "GET /resources/<uri>"
+            }
+        },
+        "service_area": {
+            "reference_point": {
+                "lat": PARIS_CITY_HALL.latitude,
+                "lon": PARIS_CITY_HALL.longitude,
+                "label": "Paris City Hall (Hôtel de Ville)"
+            },
+            "max_radius_meters": PARIS_SERVICE_AREA_MAX_METERS,
+            "bounding_box": {
+                "lat_min": 48.7,
+                "lat_max": 49.0,
+                "lon_min": 2.0,
+                "lon_max": 2.6
+            }
+        },
+        "units": {
+            "distance": "meters",
+            "time": "seconds_unix_utc",
+            "walking_distance": "meters",
+            "coordinates": "wgs84_decimal_degrees",
+            "occupancy_rate": "ratio_0_to_1"
+        },
+        "caching": {
+            "reference_ttl_seconds": REFERENCE_CACHE_TTL_MINUTES * 60,
+            "realtime_ttl_seconds": REALTIME_CACHE_TTL_MINUTES * 60,
+            "freshness_thresholds_minutes": {
+                "fresh": "<10",
+                "recent": "10-30",
+                "stale": "30-120",
+                "very_stale": ">120"
+            }
+        },
+        "enums": enums(),
+        "tools": tools_with_schemas(),
+        "resources": resources(),
+        "error_codes": error_codes()
+    })
+}
+
+/// Render the API description as Markdown. Uses [`api_description`] as its
+/// only data source so the two formats can never drift.
+#[must_use]
+pub fn api_description_markdown() -> String {
+    let desc = api_description();
+    let mut out = String::new();
+
+    let name = desc["server"]["name"].as_str().unwrap_or(SERVER_NAME);
+    let version = desc["server"]["version"].as_str().unwrap_or(SERVER_VERSION);
+    out.push_str(&format!("# {name} API ({version})\n\n"));
+
+    out.push_str("## Server\n\n");
+    out.push_str(&format!("- **Name**: `{name}`\n"));
+    out.push_str(&format!("- **Version**: `{version}`\n"));
+    if let Some(protocol) = desc["server"]["protocol"].as_str() {
+        out.push_str(&format!("- **MCP protocol**: `{protocol}`\n"));
+    }
+    out.push('\n');
+
+    out.push_str("## Service area\n\n");
+    let rp = &desc["service_area"]["reference_point"];
+    if let (Some(lat), Some(lon), Some(label)) =
+        (rp["lat"].as_f64(), rp["lon"].as_f64(), rp["label"].as_str())
+    {
+        out.push_str(&format!("Reference point: {label} ({lat:.4}, {lon:.4}).\n"));
+    }
+    if let Some(max_m) = desc["service_area"]["max_radius_meters"].as_f64() {
+        out.push_str(&format!("Maximum radius: {max_m} meters.\n"));
+    }
+    out.push('\n');
+
+    out.push_str("## Units\n\n");
+    if let Some(units) = desc["units"].as_object() {
+        for (k, v) in units {
+            if let Some(vs) = v.as_str() {
+                out.push_str(&format!("- `{k}`: {vs}\n"));
+            }
+        }
+    }
+    out.push('\n');
+
+    out.push_str("## Caching\n\n");
+    if let Some(ref_ttl) = desc["caching"]["reference_ttl_seconds"].as_i64() {
+        out.push_str(&format!("- Reference data TTL: {ref_ttl}s\n"));
+    }
+    if let Some(rt_ttl) = desc["caching"]["realtime_ttl_seconds"].as_i64() {
+        out.push_str(&format!("- Real-time data TTL: {rt_ttl}s\n"));
+    }
+    out.push('\n');
+
+    out.push_str("## Enums\n\n");
+    if let Some(enums) = desc["enums"].as_array() {
+        for e in enums {
+            if let Some(name) = e["name"].as_str() {
+                out.push_str(&format!("### {name}\n\n"));
+                if let Some(values) = e["values"].as_array() {
+                    for v in values {
+                        let val = v["value"].as_str().unwrap_or("");
+                        let descr = v["description"].as_str().unwrap_or("");
+                        out.push_str(&format!("- `{val}`: {descr}\n"));
+                    }
+                }
+                out.push('\n');
+            }
+        }
+    }
+
+    out.push_str("## Tools\n\n");
+    if let Some(tools) = desc["tools"].as_array() {
+        for t in tools {
+            let tname = t["name"].as_str().unwrap_or("");
+            let summary = t["summary"].as_str().unwrap_or("");
+            out.push_str(&format!("### `{tname}`\n\n"));
+            if !summary.is_empty() {
+                out.push_str(&format!("{summary}\n\n"));
+            }
+            if let Some(when) = t["when_to_use"].as_str() {
+                out.push_str(&format!("**When to use**: {when}\n\n"));
+            }
+            out.push_str("**Input schema**:\n\n");
+            out.push_str("```json\n");
+            out.push_str(&serde_json::to_string_pretty(&t["input_schema"]).unwrap_or_default());
+            out.push_str("\n```\n\n");
+        }
+    }
+
+    out.push_str("## Resources\n\n");
+    if let Some(resources) = desc["resources"].as_array() {
+        for r in resources {
+            let uri = r["uri"].as_str().unwrap_or("");
+            let rname = r["name"].as_str().unwrap_or("");
+            let rdesc = r["description"].as_str().unwrap_or("");
+            out.push_str(&format!("- `{uri}` — **{rname}**: {rdesc}\n"));
+        }
+    }
+    out.push('\n');
+
+    out.push_str("## Error codes\n\n");
+    if let Some(codes) = desc["error_codes"].as_array() {
+        for c in codes {
+            let code = c["code"].as_i64().unwrap_or(0);
+            let cname = c["name"].as_str().unwrap_or("");
+            let meaning = c["meaning"].as_str().unwrap_or("");
+            out.push_str(&format!("- `{code}` ({cname}): {meaning}\n"));
+        }
+    }
+    out.push('\n');
+
+    out
+}
+
+fn enums() -> Value {
+    json!([
+        {
+            "name": "StationStatus",
+            "description": "Operational state of a Velib station",
+            "values": [
+                {"value": "OPEN", "description": "Station accepts rentals and returns."},
+                {"value": "CLOSED", "description": "Station is out of service."},
+                {"value": "MAINTENANCE", "description": "Station is physically installed but not currently renting or returning bikes."}
+            ]
+        },
+        {
+            "name": "DataFreshness",
+            "description": "Age bucket of the real-time snapshot for a station.",
+            "values": [
+                {"value": "Fresh", "description": "Less than 10 minutes old."},
+                {"value": "Recent", "description": "Between 10 and 30 minutes old."},
+                {"value": "Stale", "description": "Between 30 and 120 minutes old."},
+                {"value": "VeryStale", "description": "More than 120 minutes old."}
+            ]
+        },
+        {
+            "name": "BikeTypeFilter",
+            "description": "Filter bikes by propulsion type.",
+            "values": [
+                {"value": "mechanical", "description": "Only mechanical (non-electric) bikes."},
+                {"value": "electric", "description": "Only electric-assist bikes."},
+                {"value": "any", "description": "Any bike type (default)."}
+            ]
+        }
+    ])
+}
+
+fn tools_with_schemas() -> Value {
+    let defs = tool_definitions();
+    let mut out = Vec::with_capacity(defs.len());
+    for def in &defs {
+        let name = def["name"].as_str().unwrap_or("").to_string();
+        let description = def["description"].as_str().unwrap_or("").to_string();
+        let input_schema = def["inputSchema"].clone();
+        let (when_to_use, output_schema, example_request, example_response, error_codes) =
+            tool_metadata(&name);
+        out.push(json!({
+            "name": name,
+            "summary": description,
+            "when_to_use": when_to_use,
+            "input_schema": input_schema,
+            "output_schema": output_schema,
+            "example_request": example_request,
+            "example_response": example_response,
+            "error_codes": error_codes
+        }));
+    }
+    Value::Array(out)
+}
+
+fn tool_metadata(name: &str) -> (Value, Value, Value, Value, Value) {
+    match name {
+        "find_nearby_stations" => (
+            json!("Given a user location, list nearby stations with distances and real-time availability."),
+            json!({
+                "type": "object",
+                "properties": {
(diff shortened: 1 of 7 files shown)


src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+                "lon": PARIS_CITY_HALL.longitude,
+                "label": "Paris City Hall (Hôtel de Ville)"
+            },
+            "max_radius_meters": PARIS_SERVICE_AREA_MAX_METERS,
+            "bounding_box": {
+                "lat_min": 48.7,
+                "lat_max": 49.0,
+                "lon_min": 2.0,
+                "lon_max": 2.6
+            }
```
DominicBurkart, line 62: Drift risk: these bounding-box literals (48.7/49.0/2.0/2.6) are duplicated from `Coordinates::is_valid_paris_metro` in `src/types.rs`. The stated goal of this PR is that docs cannot silently drift from behavior, but this is the one place they can. Either (a) promote those bounds to `pub const` in `src/types.rs` and reference them here like the TTLs and `PARIS_SERVICE_AREA_MAX_METERS`, or (b) add a drift-guard test that probes `is_valid_paris_metro` at the four reported corners.

---

src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+            "occupancy_rate": "ratio_0_to_1"
+        },
+        "caching": {
+            "reference_ttl_seconds": REFERENCE_CACHE_TTL_MINUTES * 60,
+            "realtime_ttl_seconds": REALTIME_CACHE_TTL_MINUTES * 60,
+            "freshness_thresholds_minutes": {
+                "fresh": "<10",
+                "recent": "10-30",
+                "stale": "30-120",
+                "very_stale": ">120"
```
DominicBurkart, line 78: The freshness threshold values (`"<10"`, `"10-30"`, `"30-120"`, `">120"`) are strings an LLM has to parse back out. Issue #9 asks for machine-readable self-description. Prefer structured bounds, e.g. `{"fresh": {"min_minutes": 0, "max_minutes": 10}}`. Bonus: derive them from `DataFreshness::from_age` so they round-trip via a drift-guard (probe boundary ages 9.99/10.0/29.99/30.0/...). Right now the thresholds in this file are hand-copied from `src/types.rs:143-148` with no test tying them together.

---

src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+//! `velib://api/description` resource (`resources/read`) render from the exact
+//! same JSON produced by [`api_description`].
+
+use serde_json::{json, Value};
+
+use crate::data::client::{REALTIME_CACHE_TTL_MINUTES, REFERENCE_CACHE_TTL_MINUTES};
+use crate::mcp::server::tool_definitions;
+use crate::types::{PARIS_CITY_HALL, PARIS_SERVICE_AREA_MAX_METERS};
+
+const SERVER_NAME: &str = "velib-mcp";
```
DominicBurkart, line 29: `MCP_PROTOCOL = "2024-11-05"` is defined here and presumably also by the `initialize` handler. Two sources of truth for the protocol version invites drift. Please either (1) source this from a single `const` re-exported from wherever `initialize` lives, or (2) add a test that asserts equality between the describe payload's `server.protocol` and the value returned by a real `initialize` JSON-RPC call.

---

src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+            "units",
+            "caching",
+            "enums",
+            "tools",
+            "resources",
+            "error_codes",
+        ] {
+            assert!(desc.get(key).is_some(), "missing top-level key: {key}");
+        }
+    }
```
DominicBurkart, line 470: Semantic concern on error code catalogue:

1. `-32600 invalid_request` is JSON-RPC spec for a malformed envelope. Annotating it as "also used for station_not_found" conflates an application error with a protocol error and will confuse client retry policies (a malformed envelope is never retryable; a missing station might be). Consider a dedicated server code (e.g. `-32002 station_not_found`) and having `get_station_by_code` return that.
2. `-32602 invalid_params` is described as a catch-all for "bad coordinates, outside service area, search radius too large, result limit exceeded, or validation failure" — five semantically distinct failure modes. For the LLM consumer (the actual audience per issue #9), splitting these into named subcodes or returning a structured `data` field would be far more actionable.
3. `DataSource` enum (`paris_open_data` / `cache` / `fallback`) exists in `src/types.rs` but is absent from the enums list here. Add it, or document why it's intentionally internal.

---

src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+        },
+        "service_area": {
+            "reference_point": {
+                "lat": PARIS_CITY_HALL.latitude,
+                "lon": PARIS_CITY_HALL.longitude,
+                "label": "Paris City Hall (Hôtel de Ville)"
+            },
+            "max_radius_meters": PARIS_SERVICE_AREA_MAX_METERS,
+            "bounding_box": {
+                "lat_min": 48.7,
```
DominicBurkart, line 58: `"time": "seconds_unix_utc"` is ambiguous — the server actually returns `last_update` as an RFC3339/ISO-8601 `DateTime<Utc>` string (see `src/types.rs` `RealTimeStatus::last_update: DateTime<Utc>`). "seconds_unix_utc" implies an integer epoch. This is the exact kind of "confidently wrong belief" issue #9 is trying to prevent. Please align the documented unit with the wire format.

---

src/mcp/server.rs
```diff
@@ -400,6 +343,149 @@ impl McpServer {
     }
 }
 
+/// Canonical list of tool definitions served by `tools/list` and re-used by
```
DominicBurkart, line 346: `tool_definitions()` reconstructs six `serde_json::Value` trees on every `tools/list` and every `describe_api` call. These are compile-time constants in practice. Wrap in `std::sync::LazyLock<Vec<Value>>` (or `once_cell::sync::Lazy` if MSRV predates 1.80) so repeated discovery calls are free. Low priority, but trivial and aligns with "Zero I/O" claim in the tool description.

---

src/mcp/server.rs
```diff
@@ -400,6 +343,149 @@ impl McpServer {
...
+        "content": [
+            {
+                "type": "text",
+                "text": text,
+            }
+        ]
+    }))
+}
+
 /// Deserialize `arguments` into the input type expected by `call`, invoke it,
```
DominicBurkart, line 489: Only `describe_api` sets `additionalProperties: false`. All other tools' inputSchemas silently accept unknown properties, which means an LLM that hallucinates a param gets no error. The schema-parity test therefore enforces drift only for declared properties. Consider adding `additionalProperties: false` to every tool in `tool_definitions()` (or explicitly documenting the laxity). This is arguably in scope for an issue whose point is to prevent confidently-wrong LLM usage.

---

tests/mcp_describe_tests.rs
```diff
@@ -0,0 +1,307 @@
...
+                    let raw = v["value"].as_str().unwrap();
+                    let json_value = Value::String(raw.to_string());
+                    let parsed: std::result::Result<BikeTypeFilter, _> =
+                        serde_json::from_value(json_value);
+                    assert!(
+                        parsed.is_ok(),
+                        "BikeTypeFilter documented value `{raw}` does not round-trip: {parsed:?}"
+                    );
+                }
+            }
```
DominicBurkart, line 285: The drift guard silently accepts unknown enum names with `_ => {}`. That means if someone adds a new enum to `describe.rs` but forgets to wire it into the match, the test passes. Flip the default to `panic!("enum `{name}` has no round-trip check; add one or remove from describe payload")` so the guard is actually a guard. The "skip gracefully" plan note is fine for enums that intentionally live only in the docs, but those should be opt-in via an explicit allowlist.

---

tests/mcp_describe_tests.rs
```diff
@@ -0,0 +1,307 @@
...
+        "jsonrpc": "2.0",
+        "id": 1,
+        "method": "tools/call",
+        "params": {"name": "describe_api", "arguments": {"format": "yaml"}}
+    }))
+    .await;
+    assert_eq!(status, StatusCode::OK);
+    assert!(
+        body["error"].is_object(),
+        "unsupported format must produce a JSON-RPC error"
```
DominicBurkart, line 305: `describe_api_rejects_unknown_format` only checks `body["error"].is_object()`. Please also assert the error `code` equals `-32602` (which the payload advertises for invalid_params) — otherwise the test doesn't enforce that the documented error contract is respected by the handler it's documenting.

---

src/mcp/describe.rs
```diff
@@ -0,0 +1,500 @@
...
+                },
+                "required": ["station", "found"]
+            }),
+            json!({"station_code": "16107", "include_real_time": true}),
+            json!({"station": null, "found": false}),
+            json!([-32600, -32602, -32001]),
+        ),
+        "search_stations_by_name" => (
+            json!("Search stations by (possibly fuzzy) name when the user refers to a station by label rather than coordinates."),
+            json!({
```
DominicBurkart, line 300: `tool_metadata` is an unchecked string-match on tool name. If someone adds a tool in `tool_definitions()` and forgets to add a branch here, the fallback returns all `Value::Null` — so the describe payload will serve a tool with `when_to_use: null`, `output_schema: null`, `example_*: null` and the tests still pass (schema-parity only checks `input_schema`). Add a test that asserts every tool in `tool_definitions()` has non-null metadata, or make the fallback a `panic!` in debug.

---

DominicBurkart (commented):
Overall direction is solid and well-aligned with issue #9 — single source of truth for tool schemas via `tool_definitions()`, TTLs imported from `data::client`, and drift-guard tests for enums and schema parity are all the right moves. No secrets/PII, no new deps (matches stated scope), diff is focused.

Main themes to address before promotion:

1. **Remaining drift surfaces**: bounding-box literals, `DataFreshness` thresholds, and `MCP_PROTOCOL` version are still hand-copied from elsewhere. The PR's thesis is "docs cannot silently drift"; these three gaps undercut it. Each needs either a `pub const` source or a round-trip test.
2. **LLM-consumer fit (the actual goal of #9)**: `"time": "seconds_unix_utc"` mismatches the RFC3339 wire format; freshness buckets are returned as parse-me strings instead of structured bounds; `-32600` is overloaded with `station_not_found` against JSON-RPC semantics; `-32602` is a catch-all for five distinct failure modes. These will produce exactly the "confidently wrong" behavior #9 wants to prevent.
3. **Guard-rails that don't guard**: the enum drift test's default arm is a silent skip, `tool_metadata` fallback is silent `Value::Null`, and only `describe_api` sets `additionalProperties: false`. Tighten each so a forgotten wiring fails loudly.
4. **Missing surface**: `DataSource` enum is documented in types but not exposed in `enums()`.

Nits: consider `LazyLock` for `tool_definitions()`; `describe_api_rejects_unknown_format` should assert the error code equals the advertised `-32602`; the markdown renderer could use `fmt::Write` instead of `format!` + `push_str`.

Draft status + janitor-promotion policy noted; leaving as COMMENT, not blocking.

---

DominicBurkart:

Review posted with 10 inline comments covering hand-copied drift surfaces (bbox / freshness thresholds / MCP protocol version), LLM-fit issues (`seconds_unix_utc` vs RFC3339, overloaded JSON-RPC `-32600`/`-32602` codes, string-encoded freshness buckets), silently-passing guards (enum default arm, `tool_metadata` null fallback, missing `additionalProperties: false`), and a missing `DataSource` enum. No reviewer comments to address. CI poller timed out — re-check status on the next run before promoting out of draft.

---

DominicBurkart:
## Cron triage — JOB_ID `qsAIJ`, OLDEST_ISSUE `one_track#2`

### Fingerprint
- branch: `claude/nice-rubin-nI3F7` (prior cron `nI3F7`, not this run's `qsAIJ`)
- last commit: `83ff42eda57f9502315f28c94159de434155bb4f`
- mergeable_state: `clean`, draft: `true`
- CI: 2/2 statuses green (`codecov/project` 59% +4%, `codecov/patch` 95.88%)
- 10 unresolved review threads, all authored by @DominicBurkart on 2026-04-21

### Triage of the 10 threads
**Trivial (≤10 LOC each, can land mechanically):**
1. `LazyLock` for `tool_definitions()` — `src/mcp/server.rs:346`
2. `additionalProperties: false` on every tool input schema — `src/mcp/server.rs:489`
3. Drift-guard panic instead of silent `_ => {}` — `tests/mcp_describe_tests.rs:285`
4. Assert error `code == -32602` in `describe_api_rejects_unknown_format` — `tests/mcp_describe_tests.rs:305`
5. Panic / non-null metadata test for `tool_metadata` fallback — `src/mcp/describe.rs:300`
6. Fix `"time": "seconds_unix_utc"` → RFC3339 to match `RealTimeStatus::last_update` — `src/mcp/describe.rs:58`
7. Single source of truth for `MCP_PROTOCOL = "2024-11-05"` (re-export from `initialize` site or assert equality in test) — `src/mcp/describe.rs:29`

**Design-shaped (need maintainer judgment, larger blast radius):**
8. Structured freshness thresholds (`{"min_minutes": 0, "max_minutes": 10}`) instead of stringly-typed ranges, derived from `DataFreshness::from_age` — `src/mcp/describe.rs:78`
9. Error-code catalogue: split `-32600` / `-32602` into named subcodes, return structured `data` field, add `DataSource` enum — `src/mcp/describe.rs:470`
10. Promote Paris bbox literals to `pub const` in `src/types.rs` OR add four-corner drift guard against `Coordinates::is_valid_paris_metro` — `src/mcp/describe.rs:62`

### Action taken this run
Status-only. Did not push to a different cron's branch (`nI3F7` ≠ `qsAIJ`) without owner sign-off, and the design-shaped threads (#8/#9/#10) are not safe to drive-by from a stateless CRON.

This run's per-repo infra observation: nanna-coder reports `/tmp/code-sign` returning `400 missing source`. If velib-mcp uses the same signer, follow-up cron pushes here will block the same way.

### Recommended next-cron action
- Owner picks structural direction on threads #8/#9/#10 (15-min decision call)
- Then a single follow-up cron lands all 10 in one focused commit, re-runs `cargo test` + `cargo clippy --all-targets --all-features -- -D warnings` + `cargo fmt --check`
- After CI re-greens, janitor flips to ready-for-review

Per cron policy: this PR remains draft. Planner jobs never promote.

---

DominicBurkart (commented):
Code review — `feat(mcp): add describe_api tool + velib://api/description resource`

Audience: coding agents. Focus: MCP spec, dispatch wiring, alignment with #9, drift-guard sufficiency, conflict surface with sibling PR #110.

## Scope check vs. issue #9
Aligned on the substantive asks: enum definitions (with descriptions), units, cache TTLs, per-tool schemas, JSON-default with markdown alt format, single fetch at session start. Both the tool and the resource surface are present and dispatched. Schema-parity + enum round-trip tests are the right drift guards.

The prior @DominicBurkart review (2026-04-21) already enumerated 10 items; most still stand on HEAD `83ff42e`. I am not re-filing those. New/additional findings below.

## New findings

### 1. Conflict surface with PR #110 (`claude/youthful-brahmagupta-dSz6w`)
Both PRs edit `src/mcp/server.rs`, `src/data/client.rs`, `README.md`, and `CLAUDE.md`. Concretely overlapping hunks:
- `src/data/client.rs` — #110 adds parser-test module and lat/lon panic fix; #114 promotes `REFERENCE_CACHE_TTL_MINUTES` / `REALTIME_CACHE_TTL_MINUTES` to `pub const`. Non-overlapping text, will merge cleanly.
- `src/mcp/server.rs` — #110 rewrites `handle_resource` into a shared-tail match; #114 adds a new `"velib://api/description"` arm to the **old** per-arm match. Textual conflict guaranteed on whichever lands second. The #114 arm also bypasses the shared error path (it `unwrap()`s via `Json(...).into_response()` — no error path to harmonize, so resolution is mechanical).
- `README.md` — both edit the Claude Code install/tools sections. Textual conflict.
- `CLAUDE.md` — #110 trims the multi-agent-process section; #114 is untouched on HEAD `83ff42e`. No conflict if #114 stays as-is.

Recommendation: whichever merges first, the other rebases — document this in the janitor queue. No code change needed here.

### 2. `resources/list` drift surface (new)
The whole PR thesis is "docs cannot silently drift," and the TTL / tool-schema paths are well-guarded. But the resources list is **duplicated**: `McpServer::process_jsonrpc_request` hardcodes five resources in `resources/list`, and `crate::mcp::describe::resources()` hardcodes the same five independently. There is no parity test between them. If a future PR adds a resource to only one, the describe payload will silently lie to LLM clients — exactly what #9 is trying to prevent.

Fix: either have `resources/list` render from `describe::resources()` (rename fields `mime_type` → `mimeType` via a small mapper), or add a parity test analogous to `schema_parity_between_tools_list_and_describe_payload`.

### 3. `describe_api_rejects_unknown_format` assertion is too loose
`tests/mcp_describe_tests.rs:305` asserts only `body["error"].is_object()`. The describe payload advertises `-32602` as the error code for this failure mode; the test should pin it:
```rust
assert_eq!(body["error"]["code"].as_i64(), Some(-32602));
```
Without this, the error-code catalogue can drift from runtime behavior and the drift-guard suite won't catch it.

### 4. `tool_metadata` silent-fallback makes the guard soft
`src/mcp/describe.rs:290` returns `(Null, Null, Null, Null, [])` for any unknown tool name. Combined with the schema-parity test only checking `input_schema` equality (not that `when_to_use` / `output_schema` / `example_*` are non-null), a future engineer adding a new tool via `tool_definitions()` but forgetting the `tool_metadata` arm would produce a payload with `"when_to_use": null` and tests would still pass. Either:
- Panic in `tool_metadata`'s `_ =>` arm (the documented LLM-consumer contract is "every tool has `when_to_use`"), or
- Add a test asserting `!tool["when_to_use"].is_null()` for every entry in `desc["tools"]`.

### 5. `"time": "seconds_unix_utc"` is factually wrong (re-iterating from prior review #6)
`RealTimeStatus::last_update` is `DateTime<Utc>` which serializes as RFC3339 (`"2024-01-15T12:30:00+00:00"`), confirmed by `src/data/client.rs` test `realtime_parses_valid_rfc3339_duedate`. Agents relying on this units table will parse the string as a unix epoch int and fail. This is the single most LLM-harmful item in the payload.

### 6. `additionalProperties: false` asymmetry
Only `describe_api`'s schema sets `additionalProperties: false`. An LLM that sends `{"latitude":..., "longitdue":...}` (typo) to `find_nearby_stations` will currently see the typo'd field silently accepted via `serde_json::from_value` ignoring unknowns — a "confidently wrong" failure mode #9 explicitly calls out. Add `additionalProperties: false` to all six tool schemas in `tool_definitions()`.

### 7. Minor
- `src/mcp/describe.rs:62` — bbox literals `48.7 / 49.0 / 2.0 / 2.6` shadow the same numbers inside `Coordinates::is_valid_paris_metro`. Promote to `pub const PARIS_METRO_BBOX` in `src/types.rs` or add a four-corner round-trip test against `is_valid_paris_metro`. (This is review item #10 from the prior review — still actionable.)
- `DataSource` enum is `pub` in `src/types.rs` but absent from `describe::enums()`. Either expose or drop the `#[serde(rename = ...)]` if it's not part of the wire contract.

## Security / PII / bloat
- No secrets, no credentials, no PII, no network I/O in the describe path. Good.
- No new deps. Good.
- 1,007 additions / 98 deletions; the new describe.rs file (~500 lines) is almost entirely data, which is appropriate — no bloat concern.
- Codecov 95.88% patch coverage; the 7 missing lines are markdown fallback `unwrap_or_default`/`unwrap_or("")` branches that are defensively unreachable — acceptable.

## Recommendation
COMMENT, not blocking. Draft + janitor-promotes policy respected. Items 2 (resources parity), 3 (error-code assertion), and 5 (wrong units) are the highest-ROI fixes and are each ≤10 LOC. Merge #110 first or pick a rebase order explicitly before either lands.

---

DominicBurkart:
## CRON triage — job-id `QO4Td`, 2026-04-24

**Fingerprint**: issue velib-mcp#9 → PR #114 → head `83ff42eda57f9502315f28c94159de434155bb4f`. Canonical PR; CI green (9/9); 10 unresolved OWNER review threads.

**This cycle attempted** to ship a successor draft PR on branch `claude/nice-rubin-QO4Td-issue-9` addressing 6 of the 10 threads. Both attempts (initial + minimal-scope retry, ~22 min combined) hit stream-idle timeouts without pushing any commits — root cause appears to be `cargo test --workspace` compile time on a cold build. Branch remains at the parent SHA unchanged; no duplicate PR was opened.

**Recommended decomposition for next CRON cycle** (one micro-PR per thread to side-step the workspace-compile budget; each should push after local validation of the single touched file):

1. `describe.rs`: `"time"` unit → `"rfc3339_utc"` + one assertion.
2. `describe.rs` / `server.rs`: consolidate `MCP_PROTOCOL` constant + parity test.
3. `server.rs::tool_definitions`: add `additionalProperties: false` to each schema + iteration test.
4. `tests/mcp_describe_tests.rs:285`: enum-drift `_` arm → panic.
5. `tests/mcp_describe_tests.rs:305`: assert error code `-32602` in unknown-format test.
6. `server.rs::tool_definitions`: `LazyLock` memoization (only if non-intrusive to the public signature).

**Larger follow-ups** (separate issues or later cycles): freshness-buckets struct, Paris bbox `pub const` with four-corner round-trip, `DataSource` enum in `describe::enums()`, `resources/list` ↔ `describe::resources()` parity, JSON-RPC error-code split, `describe_api` micro-bench, `tool_metadata` debug-assert panic path.

**Cross-PR conflict note**: PR #110 overlaps on `src/mcp/server.rs`, `src/data/client.rs`, `README.md`. Whichever lands second needs a rebase.

_Planner job; never promotes. `ready-for-review` label stays with the janitor._

---

Issue #67, opened 2026-04-04
kani proofs for API type conversion


DominicBurkart:
Part of #62 (add kani-based verification). Open one PR for this component. Be comprehensive so subsequent agents can swarm against CI errors.

## Location

`src/data/client.rs:239,279-283`

## Bug

Silent truncation when casting `u64 as u16` for capacity and bike count values parsed from the Velib API. If the API returns a value > 65535, the cast silently drops the high bits, producing a completely wrong count with no error or warning.

## Required Proofs

1. Prove that if the API returns a value > 65535, the `as u16` cast silently truncates — surface this as a formal counterexample demonstrating data corruption.
2. Prove that `u16::try_from(value)` (or a range check) is the correct safe alternative and that the fix correctly propagates errors for out-of-range values.
3. Prove that for all values in `0..=65535`, the cast is lossless (the proof obligation for the safe path).
4. If a maximum capacity is documented by the Velib API spec (e.g. stations have at most N docks), prove the cast is safe under that domain constraint and document the assumed bound.

## Setup

Add Kani as a dev dependency and write harnesses in a `#[cfg(kani)]` module within `src/data/client.rs`. Run with `cargo kani`.

PR #116, opened 2026-04-23
Tighten JSON-RPC, cover parsers + aggregation, refresh stale docs

DominicBurkart:
## Architecture
- **JSON-RPC 2.0 compliance**: `JsonRpcRequest.params` now defaults to `null`, so parameterless calls like `tools/list` need not include an empty object. [`src/mcp/types.rs`](../blob/claude/youthful-brahmagupta-eZig4/src/mcp/types.rs)
- **Honest health metrics**: removed the fabricated `hit_rate` from `velib://health`; the cache tracks no hits/misses, so the formula was cosmetic. Real cache sizes and data lag are retained. [`src/mcp/server.rs`](../blob/claude/youthful-brahmagupta-eZig4/src/mcp/server.rs)
- **Schema/handler drift fix**: `search_stations_by_name`'s advertised `limit` max was 50 while the handler enforced 100. Aligned the schema to reflect reality.
- **Dead-code cleanup**: replaced `HashMap<String, WebSocketClient>` (whose sole field was `#[allow(dead_code)]`) with a plain `HashSet<String>`; deleted the unused `McpToolHandler::test_connectivity` helper.
- **Refactor for testability**: extracted `aggregate_area_statistics` as a pure iterator-driven function, and moved `parse_reference_station` / `parse_realtime_status` to free functions (they never used `&self`).

## Testing
- New unit coverage for the Paris Open Data parsers against fixture JSON — happy path, missing required fields, and each `StationStatus` mapping (`OPEN`, `MAINTENANCE`, `CLOSED`). [`src/data/client.rs`](../blob/claude/youthful-brahmagupta-eZig4/src/data/client.rs)
- New unit coverage for `aggregate_area_statistics` — empty input, sum across stations, missing real-time data, closed station excluded from operational count. [`src/mcp/handlers.rs`](../blob/claude/youthful-brahmagupta-eZig4/src/mcp/handlers.rs)
- Routing regression guards: `tools/list` accepted with omitted `params`; advertised `search_stations_by_name` schema max matches handler's enforcement. [`tests/mcp_server_routing_tests.rs`](../blob/claude/youthful-brahmagupta-eZig4/tests/mcp_server_routing_tests.rs)
- Updated the ignored `velib://health` resource test to assert `hit_rate` is absent.

Library test count: 67 → 78; integration routing tests: 16 → 18.

## Clarity
- `docs/context/etat_actuel.md` declared Phase 2 as "Prêt à commencer" despite Phases 0-4 shipping long ago. Replaced with a concise status table linking into the delivered code.
- `docs/development/project_prompt.md` was an 18-line corrupted stub ending in `[... rest of the original prompt content ...]` that duplicated CLAUDE.md. Collapsed to a pointer to the authoritative CLAUDE.md.

## Test plan
- [x] `cargo test` — all suites green (78 lib + 18 routing + 11 types + 8 validation + 4 data_client + 2 smoke + 1 cache + doctests).
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [ ] CI: Test Suite, Rustfmt, Clippy, Security Audit, Feature Testing, Coverage, Build, all-checks gate.

https://claude.ai/code/session_01XkE52kb74MuBFgU4VrSTX6

---
_Generated by [Claude Code](https://claude.ai/code/session_01XkE52kb74MuBFgU4VrSTX6)_

### DIFF
--- a/src/data/client.rs
+++ b/src/data/client.rs
@@ -103,7 +103,7 @@ impl VelibDataClient {
             }
 
             for record in records {
-                if let Ok(station) = self.parse_reference_station(record) {
+                if let Ok(station) = parse_reference_station(record) {
                     all_stations.push(station);
                 }
             }
@@ -161,7 +161,7 @@ impl VelibDataClient {
             }
 
             for record in records {
-                if let Ok((station_code, status)) = self.parse_realtime_status(record) {
+                if let Ok((station_code, status)) = parse_realtime_status(record) {
                     all_status.insert(station_code, status);
                 }
             }
@@ -221,98 +221,6 @@ impl VelibDataClient {
             .find(|station| station.reference.station_code == station_code))
     }
 
-    /// Parse reference station data from API response
-    fn parse_reference_station(&self, record: &Value) -> Result<StationReference> {
-        let station_code = record["stationcode"]
-            .as_str()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing station code")))?
-            .to_string();
-
-        let name = record["name"]
-            .as_str()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing station name")))?
-            .to_string();
-
-        let capacity = record["capacity"]
-            .as_u64()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing capacity")))?
-            as u16;
-
-        // Parse coordinates from coordonnees_geo
-        let geo_point = record["coordonnees_geo"]
-            .as_object()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing geo coordinates")))?;
-
-        let latitude = geo_point["lat"]
-            .as_f64()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing latitude")))?;
-
-        let longitude = geo_point["lon"]
-            .as_f64()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing longitude")))?;
-
-        let coordinates = crate::types::Coordinates::new(latitude, longitude);
-
-        // Parse service capabilities
-        let capabilities = ServiceCapabilities {
-            accepts_credit_card: false,  // Not available in current API
-            has_charging_station: false, // Not available in current API
-            is_virtual_station: false,   // Not available in current API
-        };
-
-        Ok(StationReference {
-            station_code,
-            name,
-            coordinates,
-            capacity,
-            capabilities,
-        })
-    }
-
-    /// Parse real-time status data from API response
-    fn parse_realtime_status(&self, record: &Value) -> Result<(String, RealTimeStatus)> {
-        let station_code = record["stationcode"]
-            .as_str()
-            .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing station code")))?
-            .to_string();
-
-        let mechanical_bikes = record["mechanical"].as_u64().unwrap_or(0) as u16;
-
-        let electric_bikes = record["ebike"].as_u64().unwrap_or(0) as u16;
-
-        let available_docks = record["numdocksavailable"].as_u64().unwrap_or(0) as u16;
-
-        // Parse status
-        let status_str = record["is_installed"].as_str().unwrap_or("NON");
-
-        let status = match status_str {
-            "OUI" => {
-                let is_renting = record["is_renting"].as_str().unwrap_or("NON") == "OUI";
-                let is_returning = record["is_returning"].as_str().unwrap_or("NON") == "OUI";
-
-                if is_renting && is_returning {
-                    StationStatus::Open
-                } else {
-                    StationStatus::Maintenance
-                }
-            }
-            _ => StationStatus::Closed,
-        };
-
-        // Parse last update time
-        let default_time = Utc::now().to_rfc3339();
-        let last_update_str = record["duedate"].as_str().unwrap_or(&default_time);
-
-        let last_update = DateTime::parse_from_rfc3339(last_update_str)
-            .map_or_else(|_| Utc::now(), |dt| dt.with_timezone(&Utc));
-
-        let bikes = BikeAvailability::new(mechanical_bikes, electric_bikes);
-
-        let real_time_status = RealTimeStatus::new(bikes, available_docks, status, last_update);
-
-        Ok((station_code, real_time_status))
-    }
-
     /// Clean up expired cache entries
     pub async fn cleanup_cache(&self) {
         self.reference_cache.cleanup_expired().await;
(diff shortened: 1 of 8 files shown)


src/mcp/handlers.rs
```diff
@@ -233,60 +287,15 @@ impl McpToolHandler {
         &self,
         input: GetAreaStatisticsInput,
     ) -> Result<GetAreaStatisticsOutput> {
-        // Fetch live station data
         let mut data_client = self.data_client.write().await;
         let all_stations = data_client.get_all_stations(true).await?;
```
DominicBurkart, line 291: `GetAreaStatisticsInput::include_real_time` (defined in `src/mcp/types.rs` with `#[serde(default = "default_true")]`) is still deserialized but the handler unconditionally calls `get_all_stations(true)`. The PR fixed the `search_stations_by_name.limit` schema/handler drift and added a regression test for it -- same class of drift applies here. Either honor the flag (`get_all_stations(input.include_real_time)`) or drop the field from the input type and the advertised schema so coding agents aren't told they can turn real-time data off when they can't.

---

src/mcp/types.rs
```diff
@@ -191,6 +191,9 @@ pub struct JsonRpcRequest {
     pub jsonrpc: String,
     pub id: serde_json::Value,
     pub method: String,
+    /// JSON-RPC 2.0 permits omitting `params` entirely; default to null so
+    /// parameterless calls like `tools/list` need not include an empty object.
+    #[serde(default)]
```
DominicBurkart, line 196: Defaulting `params` to `Value::Null` is the right JSON-RPC 2.0 fix. Noting an adjacent gap unchanged by this PR: `jsonrpc` uses `#[serde(default = "default_jsonrpc")]`, so a request that omits the version string (or sends `"1.0"`) is silently accepted and answered with a `"2.0"` response. Spec §4 requires the server to reject such requests with `-32600 Invalid Request`. Not blocking for this PR (it's pre-existing and the PR's headline is "tighten JSON-RPC"), but worth a follow-up issue so the regression guard you added for omitted `params` has a counterpart here.

---

src/data/client.rs
```diff
@@ -221,108 +221,206 @@ impl VelibDataClient {
...
+    let geo_point = record["coordonnees_geo"]
+        .as_object()
+        .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing geo coordinates")))?;
+
+    let latitude = geo_point["lat"]
+        .as_f64()
+        .ok_or_else(|| Error::Internal(anyhow::anyhow!("Missing latitude")))?;
+
+    let longitude = geo_point["lon"]
+        .as_f64()
```
DominicBurkart, line 268: The refactor keeps the `as u16` casts for `capacity`, `mechanical`, `ebike`, `numdocksavailable` -- these are the exact u64→u16 silent-truncation sites called out in issue #67. The new parser tests exercise valid ranges but don't guard the truncation boundary. Since this PR is explicitly about parser coverage, consider adding one test that feeds `capacity: 70000` (or `mechanical: 70000`) and asserts either an error or the documented wrap, so the behavior at the boundary is pinned until #67 lands a proper fix.

---

src/mcp/handlers.rs
```diff
@@ -35,6 +35,60 @@ fn ensure_in_service_area(coords: &Coordinates) -> Result<()> {
...
+            total: total_bikes,
+        },
+        available_docks: total_available_docks,
+        occupancy_rate,
+    }
+}
+
 /// Find stations near a point, filtering by distance and a custom predicate,
 /// sorted by distance and truncated to `limit` results.
 ///
```
DominicBurkart, line 94: Docstring says "Stations without real-time data contribute to `total_stations`, `operational_stations` (per `is_operational`), and `total_capacity`, but their bike/dock counts are treated as 0 -- the data is genuinely absent, and invented zeros for `has_bikes`/`has_docks` would be misleading." The "invented zeros for `has_bikes`/`has_docks`" phrasing implies those fields exist on `AreaStatistics`; they don't -- the only fields are `available_bikes` (broken down) and `available_docks`. Suggest tightening to: "their bike and dock counts contribute 0 because the data is genuinely absent." Minor, but the doc is the public contract for coding agents.

---

tests/mcp_resource_tests.rs
```diff
@@ -147,20 +147,22 @@ async fn test_health_endpoint_returns_real_metrics() {
...
+    // Validate cache stats expose real sizes, not hardcoded or fabricated metrics.
     let cache_stats = &json_response["cache_stats"];
-    assert!(cache_stats["hit_rate"].is_number());
     assert!(cache_stats["entries"].is_number());
+    assert!(cache_stats["reference_cache_size"].is_number());
+    assert!(cache_stats["realtime_cache_size"].is_number());
+    // `hit_rate` was removed because the cache does not track hits/misses.
+    assert!(
+        cache_stats["hit_rate"].is_null(),
+        "hit_rate must not be synthesized"
```
DominicBurkart, line 158: `cache_stats["hit_rate"].is_null()` is satisfied both when the key is absent (index into `Value` yields `Value::Null`) and when it is present with an explicit `null`. The PR's intent is the former (field removed), so this test doesn't distinguish "hit_rate removed" from "hit_rate: null". Consider `assert!(!cache_stats.as_object().unwrap().contains_key("hit_rate"))` to pin the stronger property that the field is absent from the JSON body.

---

src/mcp/server.rs
```diff
@@ -20,17 +20,13 @@ use crate::{Error, Result};
 
 pub struct McpServer {
     tool_handler: Arc<McpToolHandler>,
-    clients: Arc<RwLock<HashMap<String, WebSocketClient>>>,
+    /// Live WebSocket client ids. Tracking them as a set (rather than a
+    /// `HashMap<String, Metadata>` with an unused metadata struct) keeps the
+    /// connection bookkeeping minimal while still allowing future extension.
```
DominicBurkart, line 25: The `HashSet<String>` of client ids is now written-only: it is inserted on connect and removed on disconnect but never read anywhere in the codebase (no broadcast, no count exposure, no per-client dispatch). The comment hedges "still allowing future extension," but an unused data structure guarded by a lock under every WS connect/disconnect is negative value today. Either (a) expose its size in `velib://health` as `active_ws_clients` (ties in naturally with this PR's health-metric cleanup), or (b) delete it entirely and re-introduce when there is a real consumer. "Future extension" is not a justification per the PR's own dead-code-cleanup rationale.

---

DominicBurkart (commented):
Focused, well-scoped PR. Headline wins: (1) `#[serde(default)]` on `JsonRpcRequest.params` is a correct JSON-RPC 2.0 fix with a matching regression test; (2) the advertised `search_stations_by_name` schema `limit` now matches the handler's `MAX_RESULT_LIMIT=100` and has a dedicated schema-vs-handler regression test; (3) parser unit tests cover happy path, missing required fields, and each `StationStatus` mapping; (4) `aggregate_area_statistics` extracted as a pure iterator function with four targeted tests (empty, sum, missing real-time, closed station); (5) `HashMap<String, WebSocketClient>` → `HashSet<String>` and `test_connectivity` removals are net negative LoC with no behavior change; (6) fabricated `hit_rate` gone from `velib://health`, and the resource test now asserts its absence.

All nine CI checks green against head `fae1f20`. No secrets, PII, or telemetry in the diff. Doc refresh (`etat_actuel.md`, `project_prompt.md` → CLAUDE.md pointer) accurately reflects current shipped state; the CLAUDE.md target exists and carries the canonical content.

Non-blocking findings inline:
- `GetAreaStatisticsInput::include_real_time` is still deserialized but unused in the handler — same schema/handler drift class this PR fixes for `search_stations_by_name`.
- `jsonrpc` field still defaults to `"2.0"`, so requests omitting/mis-versioning are silently accepted (spec §4 requires `-32600`). Pre-existing, but adjacent to this PR's JSON-RPC tightening theme.
- Parser coverage does not pin the `u64 as u16` truncation boundary flagged in #67; one boundary test would make the regression explicit until #67 lands.
- Aggregation docstring references `has_bikes`/`has_docks` fields that don't exist on `AreaStatistics`.
- `hit_rate` absence test relies on `Value::Null` semantics of missing keys; `contains_key` would pin the stronger "field absent" property.
- The WS `HashSet<String>` is now write-only — either expose it in health or delete it, per this PR's own dead-code-cleanup rationale.

Overall: ready to merge after addressing the trivial nits; the `include_real_time` drift is the most consequential follow-up but is also reasonable to defer to a separate PR given this one's scope. Leaving as COMMENT rather than REQUEST_CHANGES.

_Posted by an automated code-review agent._

---

DominicBurkart:
**Automated CRON triage (iteration 1)**

Code review posted (COMMENT, 6 findings — `GetAreaStatisticsInput::include_real_time` deserialized but unused (same schema/handler drift class the PR fixes for `search_stations_by_name.limit`); `JsonRpcRequest.jsonrpc` still `#[serde(default)] = "2.0"` silently accepts omitted/wrong-version requests (spec §4 violation); `aggregate_area_statistics` docstring references nonexistent `has_bikes`/`has_docks` fields; `clients: HashSet<String>` is write-only after dead-code removal pass). No in-scope comments to address. CI is 9/9 green with clean mergeable state.

Flipping to ready-for-review — findings are mergeable-as-is follow-ups.

_Stateless CRON orchestrator session._

---

PR #147, opened 2026-05-01
chore: maintenance pass (architecture, tests, clarity)

DominicBurkart:
Stateless maintenance pass against the three workstreams. Every change is small, scoped, and locally verified (`cargo build --tests`, `cargo test` (84 lib + 92 integration), `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --check`).

## 1. Architecture alignment & simplification

Findings flagged from comparing CLAUDE.md/README/docs against the actual code, then fixed:

- **Dead code (src/types.rs:303-311)**: `DataSource` enum was defined but referenced nowhere in the crate. Removed.
- **Dead code (src/data/retry.rs)**: `RetryableHttpClient::client()` getter had zero call sites. Removed.
- **Over-exposed helpers (src/data/retry.rs:243-258)**: `extract_retry_after_from_response` and `create_rate_limited_error` were `pub fn` but the only call site is `RetryableHttpClient::check_response` in the same module. Demoted to a single private fn and inlined the trivial `create_rate_limited_error` body at its one call site.
- **Doc/code drift (docs/api/data_analysis.md:14, :72)**: endpoint examples pointed at the legacy v1 search API (`api/records/1.0/search/?dataset=…`); production code in `src/data/client.rs:14-15` uses the v2.1 explore catalog endpoints. Updated docs to match.
- **Untestable formula (src/mcp/handlers.rs:345-358)**: `plan_bike_journey`'s confidence-score calculation was inlined inside an async method that needs the live data client to drive. Extracted to a private `journey_confidence_score` helper at `src/mcp/handlers.rs:105-122` (semantics preserved, with an explicit branch for the previously undefined `max_walk_meters == 0` case) and the handler now just calls it.

## 2. Testing posture

Highest-value gaps addressed:

- **`journey_confidence_score` (src/mcp/handlers.rs)** — *previously untested* (inlined inside `plan_bike_journey`, only reachable via integration tests against the live API). Added 5 unit tests pinning down the contract: perfect-placement → 1.0, both walks at budget → 0.5, walks beyond budget clamp at 0.1, score in `[0.1, 1.0]` for a representative spread of inputs (incl. `(u32::MAX, u32::MAX, u32::MAX)`), and the explicit zero-budget branch.
- **`Coordinates::distance_to` (src/types.rs)** — *had a unit test, no executable usage example*. Added a doctest demonstrating the canonical Paris-center → Louvre call and the `distance_to(self) == 0.0` identity, so docs.rs renders a working example.
- **Refactored bloated/redundant suites — no coverage loss**:
  - Deleted `tests/error_tests.rs` in full: every assertion was a weaker duplicate of unit tests already in `src/error.rs` (`mcp_error_code_invalid_params_variants`, `mcp_error_code_internal_variants`, `error_type_covers_all_non_http_variants`, `rate_limited_display_with_retry_after`, `rate_limited_display_without_retry_after`).
  - Trimmed `tests/mcp_server_dispatch_tests.rs` from 8 tests to 1: 7 were weaker variants of `tests/mcp_server_routing_tests.rs` cases (`tools_list`, `resources_list`, `unknown_method`, `unknown_tool`, `malformed_json`, `missing-tool-name`, request-id preservation); kept only `health_endpoint_returns_healthy`, which is the one case that exercises `Server::router()` rather than `McpServer::router()`.

Net: +6 unit tests + 1 doctest, ~300 lines of duplicate test code removed.

## 3. Clarity

- **README.md:19, :57, :71, :82** — replaced four occurrences of misleading `--port` CLI flags. The binary has no argv parsing; it reads `PORT`/`IP` from env (`src/server/config.rs::parse_server_address`). Updated examples to use the env-var form. Moved the inline JSON for the Cursor section out of a bash fence so editors render it correctly.
- **CLAUDE.md** — dropped the "Métriques de Performance / Gains Attendus" block (`-40%` / `-60%` / `+25%` / `+50%`); the section header explicitly labels the numbers as *expected* with no measurement or source, so presenting them as facts misleads contributors. Replaced the brittle "Suite de tests complète (18+ tests)" with a phrasing that doesn't go stale every time a test is added. Stripped leading slashes from repo-relative file pointers in "Fichiers Importants" (`/src/main.rs` → `src/main.rs`) so they match the file:line pointer style used in PR / commit messages.

## Notes

- Local commits were authored with conventional prefixes (`arch:` / `test:` / `clarity:`); the GitHub MCP push routes each modified file through a separate API commit, so the branch shows one GitHub-side commit per file rather than the four logical commits I authored locally. The commit messages on each preserve the intent of the originating local commit.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L3gFY2LhHNQWKKBUXHTuVp)_


src/mcp/handlers.rs
```diff
@@ -89,6 +89,38 @@ where
...
+        } else {
+            0.1
+        };
+    }
+    let max_walk = f64::from(max_walk_meters);
+    let pickup_ratio = f64::from(pickup_distance_meters) / max_walk;
+    let dropoff_ratio = f64::from(dropoff_distance_meters) / max_walk;
+    let score = 1.0 - f64::midpoint(pickup_ratio, dropoff_ratio) * 0.5;
+    score.clamp(0.1, 1.0)
+}
```
DominicBurkart, line 122: Behavior change worth flagging: the previous inline implementation (when `max_walk_meters == 0`) computed `f64::from(d) / 0.0` which yields `+inf` (or `NaN` for `0/0`), then `1.0 - inf*0.5` becomes `-inf`/`NaN`, finally `clamp(0.1, 1.0)` produced `0.1` for any non-zero distance and `NaN.clamp(...)` returns `NaN` for the `(0, 0, 0)` case. The new explicit branch returns `1.0` for `(0, 0, 0)` instead of `NaN`. That's clearly an improvement, but it is a semantic change, not a pure refactor — please call it out in the PR body so a future bisect lands here cleanly. Otherwise the extracted helper + tests are exactly the kind of unit-testable seam this codebase needed.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

src/data/retry.rs
```diff
@@ -240,23 +240,15 @@ impl Default for RetryPolicy {
     }
 }
 
-/// Helper function to extract retry-after header from reqwest error
-pub fn extract_retry_after_from_response(response: &reqwest::Response) -> Option<u64> {
+/// Extract the `Retry-After` header (seconds) from a response, if present and parseable.
+fn extract_retry_after_from_response(response: &reqwest::Response) -> Option<u64> {
```
DominicBurkart, line 244: Demoting the helper to private and inlining the trivial `create_rate_limited_error` is justified by the precept of removing over-exposed surface area, and `cargo search` confirms zero external call sites in this repo. Two notes for future maintenance:

1. Removing `pub fn extract_retry_after_from_response` and `pub fn create_rate_limited_error` is a SemVer-breaking change for any downstream that was using `velib-mcp` as a library. Looking at `Cargo.toml`/registry, this crate appears to only be consumed as a binary today, so the risk is theoretical — but worth a one-line note in the changelog/PR body.
2. The new doc comment on `extract_retry_after_from_response` ("seconds") is slightly misleading: the `Retry-After` header may also be an HTTP-date per RFC 9110 §10.2.3, in which case `parse::<u64>()` will silently fail and return `None`. Today this is the same behavior as before, but the new doc reads as if seconds is the only valid form. Consider either widening the doc to mention HTTP-date is intentionally not handled, or filing an issue. Not blocking.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

tests/error_tests.rs
```diff
@@ -1,119 +0,0 @@
-use velib_mcp::Error;
```
DominicBurkart, line 1: Spot-checked the dedup claim against `src/error.rs`: `mcp_error_code_invalid_params_variants`, `mcp_error_code_internal_variants`, `error_type_covers_all_non_http_variants`, and the two `rate_limited_display_*` cases do indeed cover the same ground these integration tests were exercising — and the unit tests use the variant-name-as-source-of-truth pattern (compiler-enforced) rather than re-listing every variant in a `Vec<Error>` literal. The integration variants would silently rot the next time someone added an `Error` variant. Deletion is the right call.

One small ask: please mention in the PR body that the `cargo deny`/`cargo audit` posture is unchanged (these were not security tests), so a security-conscious reviewer doesn't have to reverify.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

tests/mcp_server_dispatch_tests.rs
```diff
@@ -1,196 +1,21 @@
...
-
-    assert_eq!(body["id"], "my-string-id");
-}
 
 #[tokio::test]
 async fn health_endpoint_returns_healthy() {
-    // The /health route is on Server, not McpServer. We need the full server router.
     let addr: std::net::SocketAddr = "127.0.0.1:0".parse().unwrap();
     let server = velib_mcp::Server::new(addr);
     let router = server.router();
```
DominicBurkart, line 21: Verified against `tests/mcp_server_routing_tests.rs`: the routing file already covers tools/list (asserts all five names AND `tools.len() == 5`, stricter than the deleted `>= 5`), resources/list (asserts all four URIs AND `resources.len() == 4`), unknown_method, unknown_tool, malformed JSON (`code == -32700`), missing-tool-name (500 + body), and string-id preservation. So the dedup is real and the kept `health_endpoint_returns_healthy` is the unique case (it goes through `Server::router()` rather than `McpServer::router()`).

Suggestion (non-blocking): with only one test left, consider renaming this file to `tests/server_health_test.rs` (or moving the test into the existing `tests/mcp_server_routing_tests.rs` with a comment noting it uses `Server::router()`). The file's docstring already explains the split, but the filename `mcp_server_dispatch_tests.rs` no longer matches the contents.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

CLAUDE.md
```diff
@@ -210,14 +210,6 @@ done
 - [ ] Documentation à jour
 ```
 
-### Métriques de Performance
-
-**Gains Attendus vs Processus Linéaire**
-- **Temps cycle**: -40% (parallélisation phases)
-- **Temps attente**: -60% (élimination handoffs)
-- **Qualité code**: +25% (validation continue)
```
DominicBurkart, line 218: Good removal — the `-40% / -60% / +25% / +50%` numbers were unsourced and the surrounding section explicitly labeled them "expected", which makes them aspirational but presents them as factual. For an audience of coding agents these are exactly the kind of "confident-looking number with no provenance" hooks that get re-emitted verbatim in PR descriptions and slowly become folk-truth. Trimming this is the right call against the bloat-risk precept.

Tangentially: while you're in CLAUDE.md, the surrounding "Phases Terminées" list (lines 35-41) and "Suite de tests complète (18+ tests)" had the same staleness vector. The latter you fixed; the former still ages every time a new phase ships. Out of scope for this PR — flagging only so it lands on someone's mental TODO.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

README.md
```diff
@@ -14,9 +14,9 @@ A high-performance Model Context Protocol (MCP) server providing access to Paris
 Install and use the Velib MCP server with Claude Code in one command:
 
 ```bash
-# Install and configure the server
+# Install and configure the server (port is read from the PORT env var; default 8080)
 cargo install --git https://github.com/dominicburkart/velib-mcp.git
-claude config add-server velib-mcp "cargo run --release -- --port 3000"
+claude config add-server velib-mcp "PORT=3000 velib-mcp"
```
DominicBurkart, line 19: Confirmed against `src/server/config.rs::parse_server_address` and `src/main.rs`: there is no argv parsing — only `PORT` and `IP` env vars. Every prior `--port` reference would have silently failed (or been ignored if `clap` were missing). Fix is correct and important.

Minor: `claude config add-server velib-mcp "PORT=3000 velib-mcp"` works in a POSIX shell because of word-splitting, but Claude Code's add-server takes the command string and may not invoke it through a shell on all platforms (e.g. Windows). For maximum portability, the canonical form is `claude mcp add velib-mcp --env PORT=3000 -- velib-mcp` (or the JSON-config equivalent shown in the Cursor section below). Worth verifying empirically before this README guides a user into a misconfiguration. Non-blocking — the previous `--port 3000` was strictly worse.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

docs/api/data_analysis.md
```diff
@@ -11,7 +11,7 @@ Ce document analyse les deux jeux de données Velib disponibles via l'API Open D
 
 ### Endpoint API
 ```
-https://opendata.paris.fr/api/records/1.0/search/?dataset=velib-disponibilite-en-temps-reel
+https://opendata.paris.fr/api/explore/v2.1/catalog/datasets/velib-disponibilite-en-temps-reel/records
```
DominicBurkart, line 14: Verified against `src/data/client.rs` — production code does use the v2.1 `explore/v2.1/catalog/datasets/.../records` endpoints, so this fix is correct and meaningful (doc-driven onboarding would otherwise send a contributor down the v1-search path that no longer matches the client's URL builder). Good catch.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

src/types.rs
```diff
@@ -300,16 +315,6 @@ impl VelibStation {
...
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub enum DataSource {
-    #[serde(rename = "paris_open_data")]
-    ParisOpenData,
-    #[serde(rename = "cache")]
-    Cache,
-    #[serde(rename = "fallback")]
-    Fallback,
-}
```
DominicBurkart, line 311: Confirmed `DataSource` has no in-crate references (only mentioned descriptively in `docs/api/mcp_schemas.md`, not imported anywhere). Removal is correct.

Caveat: this is a public type in a public module — removing it is technically a SemVer-breaking change. As with the `retry.rs` helpers, this matters only if anything consumes `velib-mcp` as a library, which today appears to be no. Worth one bullet in the PR body so the impact is searchable. The `Cache` and `Fallback` variants in particular suggest there was once an architectural intent (response provenance tagging) that was never wired through — if that intent is still alive, archive it as an issue rather than just deleting; if it's dead, the deletion is unambiguously correct.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

DominicBurkart:
**Cron status — 3 subagents complete**

CR: prior CRON CR already covered head `6caa6ca` with 8 inline comments (skip — already reviewed).

Address-comments evaluated all 8 inline threads — 4 explicitly self-labeled "non-blocking" / "out of scope" / "flagging only", 1 pure approval, 3 request PR-body text annotations rather than code changes. **Deferred** as nits-only on a green PR (judgment call per directive).

CI: GREEN — all 10 checks passing, `mergeable_state: clean`, `cargo deny` + codecov + clippy + tests + cargo-sort + license-compliance + security-audit all green. **Flipped draft → ready for review.**

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

